PR #23, opened 2014-05-30
[scripts] add travis-build.sh and .travis.yml file

eren:
Github provides travis continuous integration service. This change makes it possible to utilize travis with prebuilt toolchains from http://newos.org/toolchains/

The build matrix is defined in .travis.yml file. I don't know to full extend which toolchain should be used for each project. This list need to be extended. If you can provide me with build matrix, I can update the PR.

The sample output for travis can be seen in https://travis-ci.org/eren/lk

Regards,
Eren


travisg:
Hmm, finally looking at this. Will take and update on the master side.

travisg:
taken upstream, thanks!

PR #26, opened 2014-10-22
some bug fix and improvement

ghost:
1.fix the build break
2.implement fiq function
3.remove 32bit code


travisg:
Closing this after far too long, and the delay here was on the project's side, not yours — Travis said in 2014 he'd work through it commit by commit, you followed up with a fix for the `device_find` loop, and then it went quiet from us. Sorry about that.

Ten years on there's no honest way to pick this up as a branch: it's 73 files against a tree that has since been through the spinlock rework Travis mentioned at the time, plus SMP, the VM rewrite, and the module system. Nothing in it rebases.

If any of the individual fixes still bite you on current master, please do open them as small separate PRs — those would be looked at promptly now.

PR #30, opened 2014-12-10
[make] add support for GCC 4.9 diagnostic colors

M1cha:


travisg:
Thanks, I made a more or less equivalent patch without the test. It seems if you just set GCC_COLORS to something nonzero it'll default to auto on the colorized output, which is good enough.

PR #32, opened 2014-12-12
X86 64

srodrig1:
Check if that can be portable enough there is no hash defines regarding architecture, I,m using __alignof__ to align every new declared data structure to the minimal aligned required from the proper architecture, I created a new macro for that, let me know if you think this solve the problem, or maybe look for other solutions.


travisg:
I just did a big merge of srodrig1/master into master here, so I think this is part of that change? Going to close. Reopen if I missed something.

PR #36, opened 2015-02-03
[kernel] thread_join from multiple threads

jaybh:
This patch enables multiple concurrent threads to call thread_join on
a particular thread

In thread_join(), we simply make checks regarding whether the thread
structure has been freed or not.

This could just be a quick fix. A better solution would be to
implement reference counting on thread objects, and free the
resources when the count goes to zero.


travisg:
Filed a bug and closed it out after.... 12 years? wow. Yeah, this isn't a solid solution short of having some sort of handle based scheme.

PR #40, opened 2015-03-06
[vm] search translation table to for phys2virt

M1cha:
this allows us to correctly translate addresses that were not mapped
by the initial table.


travisg:
What is this trying to solve? I cannot figure out what it's doing.

M1cha:
if u try to the virtual address from a physical one that got mapped dynamically using arch_mmu_map you won't get a result using the current way.

PR #41, opened 2015-03-06
OpenRISC 1000 (or1k) architecture port and or1ksim platform

skristiansson:
This pull request adds support for the OpenRISC 1000 architecture and an or1ksim platform that uses the architecture.

OpenRISC is an open cpu architecture project, completely community driven and it has several
open source hardware implementations as well as simulator implementations.
The port presented here is aimed at implementations that are compliant with version 1.1 of the architecture
specification:
https://github.com/openrisc/doc/blob/master/openrisc-arch-1.1-rev0.pdf?raw=true

The or1ksim "golden reference" simulator is one of the implementations that fulfills that requirement,
and that has been chosen as the supporting platform.
https://github.com/openrisc/or1ksim
Building the simulator should be fairly straight forward, a simple
'./configure --target=or1k-elf && make && make install' should do.

Apart from running the or1ksim platform in or1ksim, most existing other platforms matches
that in terms of memory map for the UART etc.
I have also tested the port by running it on mor1kx, which is a hardware implementation:
https://github.com/openrisc/mor1kx
and jor1k, an or1k simulator implemented in javascript:
https://github.com/s-macke/jor1k
A live demo of that can be found here:
http://oompa.chokladfabriken.org/openrisc/jor1k/demos/lk.html

Instructions on how to build an or1k-elf- toolchain needed for this port can be found here:
http://opencores.org/or1k/OpenRISC_GNU_tool_chain#Newlib_toolchain_.28or1k-elf.29

A few final notes about the port
- MMU is supported, but can be opted out by not defining WITH_KERNEL_VM on a platform level.
- The uart code in the or1ksim platform is close to a 1:1 copy of the omap3 uart, which both are ns16550 uarts. It could be a good idea to break this out into a generic driver instead.
- As things stands today, the cpu never leaves kernel mode.  
Mostly because I couldn't find any references in the other ports of initiating usermode threads.  
So, I'd like to get some input on that topic.  
In terms of other changes needed to support that would basically be to switch to a kernel stack on exception entry from usermode, currently the exception handlers use the stack that got interrupted.

### DIFF
--- a/arch/or1k/arch.c
+++ b/arch/or1k/arch.c
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2015 Travis Geiselbrecht
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files
+ * (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge,
+ * publish, distribute, sublicense, and/or sell copies of the Software,
+ * and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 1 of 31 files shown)


travisg:
Super cool! I'll give it a whirl on the emulator tonight, but see no reason why I shouldn't be able to take it as is.

travisg:
Neat! I built a toolchain and was able to run it. Works pretty well, though the time seems to be somewhat off in the simulator.

As far as user space, at the moment lk never enters user space. It's primarily used in kernel mode only embedded systems, though I'm thinking about adding an optional add-on user space that sits on top of the VMM.

skristiansson:
Yeah, or1ksim's timer doesn't really reflect real time (and not really simulated clock cycles neither).
When running on a real target, at least the clock_tests count to 5 takes 5 seconds,
so it seems to be alright.

Thanks for the info about user space, that was kind of what I've figured, but better ask than sorry ;)
I'll keep an eye out for your potential updates in that field though.

Also, thanks for merging.

PR #42, opened 2015-03-07
[vm] allow using null pointers

M1cha:
on devices where SDRAM starts at 0x0 null pointers were treated as a error.
To work around this problem, 'pmm_alloc_kpages' now returns a result code
and stores the allocated address in a void** pointer.


M1cha:
it turns out this was just a workaround for a wrong MMU mapping

PR #43, opened 2015-03-09
[arch][arm] fix disabling WITH_KERNEL_VM

M1cha:
inspired by the commit from @xiaoxiang-xiaomi but without disabling the MMU

### DIFF
--- a/arch/arm/arm/arch.c
+++ b/arch/arm/arm/arch.c
@@ -141,7 +141,7 @@ void arch_chain_load(void *entry, ulong arg0, ulong arg1, ulong arg2, ulong arg3
 
 	arch_quiesce();
 
-#if WITH_KERNEL_VM
+#if ARM_WITH_MMU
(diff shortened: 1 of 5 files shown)


zeusk:
what's the point of using mmu without vm ?

travisg:
You have to use it to at least set up an identity map with different cache
settings on main memory and device memory before you can enable the data
cache. That's what the old mmu code on lk used to do prior to the VM. It
could only handle 1MB segments.

M1cha:
didn't the novm commits fix this?

travisg:
It might have. I have been thinking of revisiting this soon. There is some new mmu stuff coming down the pipe soon, to fully support user space, and I figured I'd look at it then.

M1cha:
I'm not sure if this was fixed on main, but given how much LK evolved this change probably doesn't make sense anymore.

PR #50, opened 2015-07-02
update CMSIS for arm-m

cpu-chromium:
1- License has changed, it looks more permissive now.
2- I did not try to preserve the changes, the files have 5 years worth of differences.

This also includes a slightly better readme.md which I can drop.


travisg:
This is already upstream, right?

cpu-chromium:
Yep, I think this has been merged.

travisg:
Okay, closing.

PR #57, opened 2015-10-09
[spiflash][stm32f7] Protect SPIFlash driver with a mutex.

gkalsi:
Spiflash operations should be atomic. Guard external qspi methods with a mutex.


travisg:
LGTM. push it

PR #62, opened 2015-10-23
_Static_assert is called static_assert in C++

ErikCorryGoogle:


travisg:
Thanks, committed and pushed.

PR #63, opened 2015-10-27
Cmpctmalloc

ErikCorryGoogle:
Cmpctmalloc, third malloc implementation for lk

See https://codereview.appspot.com/268650043 and comment near the top of cmpctmalloc.c

This is a simple O(1) alloc/free with eager coalescing of free areas and no per-thread arenas or other memory that is hoarded and not freed up.   This version does not return memory to the OS, but that would be easy to add.  There is a global mutex.


lib/heap/cmpctmalloc/cmpctmalloc.c
```diff
@@ -0,0 +1,608 @@
...
+	if (payload == NULL) return cmpct_alloc(size);
+	header_t *header = (header_t *)payload - 1;
+	size_t old_size = header->size - sizeof(header_t);
+	void *new_payload = cmpct_alloc(size);
+	memcpy(new_payload, payload, MIN(size, old_size));
+	cmpct_free(payload);
+	return new_payload;
+}
+
+void cmpct_delayed_free(void *ptr)
```
travisg: The delayed free stuff is all up in lib/heap/heap_wrapper.c now, so all of that can be removed out of this file. Far as I can tell no one calls this anyway.

arch/arm64/toolchain.mk
```diff
@@ -8,10 +8,14 @@ ifeq ($(FOUNDTOOL),)
 ARCH_arm64_TOOLCHAIN_PREFIX := aarch64-linux-android-
 FOUNDTOOL=$(shell which $(ARCH_arm64_TOOLCHAIN_PREFIX)gcc)
 ifeq ($(FOUNDTOOL),)
+ARCH_arm64_TOOLCHAIN_PREFIX := aarch64-linux-gnu-
```
travisg: doesn't belong in this changelist, but it could go in somewhere else.

lib/heap/cmpctmalloc/cmpctmalloc.c
```diff
@@ -0,0 +1,608 @@
...
+#if WITH_KERNEL_VM && !defined(HEAP_GROW_SIZE)
+#define HEAP_GROW_SIZE (4 * 1024 * 1024) /* Grow aggressively */
+#elif !defined(HEAP_GROW_SIZE)
+#define HEAP_GROW_SIZE (4 * 1024) /* Grow less aggressively */
+#endif
+
+STATIC_ASSERT(IS_PAGE_ALIGNED(HEAP_GROW_SIZE));
+
+// Enough in practice even for 64 bit CPUs.
+#define HEAP_ALLOC_VIRTUAL_BITS 48
```
travisg: May be nice to trim this down for 32bit cpus, and maybe allow something in the build system to reduce it farther if they know the max size of the heap wont exceed a certain value.

scripts/do-qemuarm
```diff
@@ -18,17 +20,20 @@ DO_NET_TAP=0
...
+while getopts bdhm:cMnt6 FLAG; do
     case $FLAG in
         b) DO_BLOCK=1;;
+        c) DO_CMPCTMALLOC=1;;
         d) DO_DISPLAY=1;;
+        M) DO_MINIHEAP=1;;
         n) DO_NET=1;;
         t) DO_NET_TAP=1;;
         6) DO_64BIT=1;;
-        6) DO_MEM=1;;
```
travisg: nice copy-paste bug there!

travisg:
Generally looks pretty good. It'll be interesting to see how this one performs both time/space wise. Will also need a trim routine pretty soon.

lib/heap/cmpctmalloc/cmpctmalloc.c
ErikCorryGoogle: Deleted

lib/heap/cmpctmalloc/cmpctmalloc.c
```diff
@@ -0,0 +1,723 @@
...
+		set_free_list_bit(index);
+		bucket = &theheap.free_lists[index];
+	}
+	free_t *old_head = *bucket;
+	if (old_head != NULL) old_head->prev = free_area;
+	free_area->next = old_head;
+	free_area->prev = NULL;
+	*bucket = free_area;
+	theheap.remaining += size;
+#ifdef DEBUG
```
travisg: Might be nice to move this to some sort of local #define, like HEAP_DEBUG then can be set at the top of the file.

travisg:
After running this a bit I'm wondering if there's a bug in the trim code. If I load it up in qemu arm32 and run 'fibo 22' which should allocate a ton of 4K pages + some 200something byte structures (stacks and threads) and freeing them, then heap trimming, I see a lot of:

	bucket 53
		base 0x88873008, end 0x88873ff8, len 0xff0
		base 0x8bc80008, end 0x8bc80ff8, len 0xff0
                ...

Seems like those are perfect things to be returning pages to the system.

travisg:
Thanks! Took this and merged it upstream.

PR #64, opened 2015-10-27
This is floating point support for the x86 and x86-64 arch

srodrig1:


travisg:
Looks pretty good, I'll take this upstream.

travisg:
Okay, taken upstream. Thanks!

PR #65, opened 2015-10-27
Fix carriage return handling on the PC platform

gabemblack:


travisg:
Looks pretty good, I'll merge it here in a little bit.

travisg:
merged

PR #67, opened 2015-11-03
Block IO Aligned Buffers

gkalsi:
+ Add a flags field inside `bdev_t`
+ Allow devices to mark themselves as requiring cache aligned buffers
+ [STM32F7 Spiflash] require cache aligned buffers
+ [STM32F7 Spiflash] use default implementation for `bio_read`
+ Don't allocate on the stack for `bio dump` since this blows the stack

### DIFF
--- a/lib/bio/bio.c
+++ b/lib/bio/bio.c
@@ -401,7 +401,8 @@ void bio_initialize_bdev(bdev_t *dev,
                          size_t block_size,
                          bnum_t block_count,
                          size_t geometry_count,
-                         const bio_erase_geometry_info_t* geometry)
+                         const bio_erase_geometry_info_t* geometry,
+                         const uint8_t flags)
 {
     DEBUG_ASSERT(dev);
     DEBUG_ASSERT(name);
@@ -419,6 +420,7 @@ void bio_initialize_bdev(bdev_t *dev,
     dev->geometry = geometry;
     dev->erase_byte = 0;
     dev->ref = 0;
+    dev->flags = flags;
 
 #if DEBUG
     // If we have been supplied information about our erase geometry, sanity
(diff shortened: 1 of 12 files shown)


lib/bio/bio.c
```diff
@@ -51,6 +51,11 @@ static ssize_t bio_default_read(struct bdev *dev, void *_buf, off_t offset, size
     int err = 0;
     STACKBUF_DMA_ALIGN(temp, dev->block_size); // temporary buffer for partial block transfers
```
gkalsi: While I'm in here, does this need to be aligned to a `block_size` or can we get away with `CACHE_LINE` ?

lib/bio/bio.c
```diff
@@ -75,22 +75,34 @@ static ssize_t bio_default_read(struct bdev *dev, void *_buf, off_t offset, size
...
-    if (len >= dev->block_size) {
-        /* do the middle reads */
-        size_t block_count = len / dev->block_size;
-        err = bio_read_block(dev, buf, block, block_count);
+    if (requires_alignment) {
+        while (len >= dev->block_size) {
+            /* do the middle reads */
+            err = bio_read_block(dev, temp, block, 1);
+            memcpy(buf, temp, dev->block_size);
+            if (err < 0)
```
travisg: should double check that < 0 is sufficient. It may be that bio_read_block returns 0, which should probably be considered in the bail path.

travisg:
looks pretty good, with a few nits and assuming it passes tests.

lib/bio/bio.c
gkalsi: good call, we should bail if no bytes are read, will fix.

gkalsi: Actually on second thought we should bail if `err != dev->block_size` ... using that instead

travisg:
LGTM

gkalsi:
Merged.

PR #70, opened 2015-11-09
[arch][x86][fpu]Change naming convention for FPU flag

bzhu5:
To align with lk/arm flag naming convention, FPU flag
ENABLE_FPU is changed to X86_WITH_FPU


travisg:
done. merged.

PR #72, opened 2015-11-11
[arch][x86_64][fpu]use correct initial fpu states when each thread is created

bzhu5:
Currently, all new thread created with **zero** fpu states (512byte buffer of fxsave area is filled up with 0s), then when a new thread triggers "fpu device not available" exception (lazy fpu algorithm) at the first time, the handler just blindly call fxrstor to fetch all those zero states from fxsave area buffer. This patch is just to guarantee each new thread to have correct initial fpu states. 

Note that x86 fpu code isn't changed in this patch because the 32/64 bit fpu code are 99% duplicated. I assume that they should be merged in future.

BTW, this is not fixing https://github.com/travisg/lk/issues/71

### DIFF
--- a/arch/x86-64/fpu.c
+++ b/arch/x86-64/fpu.c
@@ -48,6 +48,9 @@
 static int fp_supported;
 static thread_t *fp_owner;
 
+/* FXSAVE area comprises 512 bytes starting with 16-byte aligned */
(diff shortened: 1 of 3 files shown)


travisg:
merged into master

Issue #73, opened 2015-11-11
new ioctl like BIO_IOCTL_GET_MEM_MAP with no side effects


cpu-chromium:
Issuing ioctl with BIO_IOCTL_GET_MEM_MAP for qspi causes it to enter into linear mode, it would be nice to have an ioctl which gives me the address it would be mapped to but without causing it to enter linear mode, so I can get the address it would map.

gkalsi:
Closing. Fixed by #74

PR #74, opened 2015-11-11
[bio][flash] Added bio ioctl to return memory mapped address without …

gkalsi:
…putting the device into linear mode


travisg:
LGTM

gkalsi:
Closing this. Fixes #73

PR #75, opened 2015-11-12
cmpctmalloc: Tweak bucket sizes to optimize power-of-2 allocs.

ErikCorryGoogle:
With this change allocations of > 128 bytes that are powers of two will be allocated at the exact size requested, instead of being subjected to worst-case round-up of 1/8th (12.5%). 

Fix and test heap_trim better.  The tests now do trims with tricky free sizes at the start and end of OS allocation areas (just around 1 page).

Add an assert for the most obvious double frees in debug mode.


travisg:
merged

PR #79, opened 2015-11-24
[arch][x86][x64][fpu]fix compile error when X86_WITH_FPU is not defined.

bzhu5:
With this patch, no compile failure issue when either X86_WITH_FPU not defined
or defined as 0(1).


travisg:
thanks, committed

PR #80, opened 2015-12-03
[qemu][cortex-m4] Support qemu-m4 target.

gkalsi:


travisg:
LGTM, flatten and merge

PR #81, opened 2015-12-07
Pr/memorylcd

gkalsi:


target/dartuinoP0/memory_lcd.c
```diff
@@ -0,0 +1,178 @@
...
+    return NO_ERROR;
+}
+
+
+
+static void mlcd_flush(uint starty, uint endy)
+{
+    chip_select(true);
+
+    uint8_t localbuf[MLCD_BUF_SIZE];
```
travisg: size on stack.

travisg:
looks pretty good except maybe the relatively large size of that buffer. Might want to make it static.

gkalsi:
Squashed and merged into master.

PR #82, opened 2015-12-08
[gfx][colour] Separate surface manipulation functions from colour tra…

gkalsi:
…nslation functions.

Reduce code duplication in the future. `copyrect`, `fillrect`, and `putpixel` are identical regardless of the colour representation (8bit rgb vs 8bit mono). Shifting the responsibility of flattening colours away from these functions.

### DIFF
--- a/include/lib/gfx.h
+++ b/include/lib/gfx.h
@@ -63,6 +63,7 @@ typedef struct gfx_surface {
 	uint alpha;
 
 	// function pointers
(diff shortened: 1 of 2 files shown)


travisg:
LGTM

travisg:
Also I saw what you did there.

gkalsi:
Did I do something clever without realizing it? :stuck_out_tongue:

PR #83, opened 2015-12-10
[fs][spifs] ioctl for getting memmap addr of a file

gkalsi:


lib/fs/debug.c
```diff
@@ -78,6 +78,47 @@ STATIC_COMMAND_END(fs);
...
+            if (err != NO_ERROR) {
+                fs_close_file(handle);
+                return err;
+            }
+
+            printf("%s is mapped at %p\n", argv[3].str, file_addr);
+
+            return fs_close_file(handle);
+        }
+    }
```
nvll: Might be good to add the break now so it isn't easily missed when it's time to add another case, or a default.

lib/fs/debug.c
```diff
@@ -126,10 +168,10 @@ static int cmd_fs(int argc, const cmd_args *argv)
...
 
         err = fs_format_device(
-                argv[2].str,
-                (argc >= 4) ? argv[3].str : NULL,
-                NULL
-        );
+                  argv[2].str,
+                  (argc >= 4) ? argv[3].str : NULL,
+                  NULL
+              );
```
nvll: Is this a github problem? This looks funky.

lib/fs/spifs/spifs.c
```diff
@@ -1139,6 +1139,42 @@ static status_t spifs_fs_stat(fscookie *cookie, struct fs_stat *stat)
...
+    LTRACEF("cookie %p, argp %p\n", cookie, argp);
+
+    status_t result;
+
+    spifs_file_t *file = (spifs_file_t *)cookie;
+    spifs_t *spifs = file->fs_handle;
+    bdev_t *dev = spifs->dev;
+
+    // Get the base address of the underlying BIO device.
+    void* result_addr;
```
nvll: void* vs void * is inconsistent with your other pointers above.

lib/fs/spifs/spifs.c
```diff
@@ -1139,6 +1139,42 @@ static status_t spifs_fs_stat(fscookie *cookie, struct fs_stat *stat)
     return NO_ERROR;
 }
 
+static status_t spifs_ioctl_get_file_addr(filecookie *cookie, void** argp)
```
nvll: void** vs void **

lib/fs/spifs/spifs.c
```diff
@@ -1139,6 +1139,42 @@ static status_t spifs_fs_stat(fscookie *cookie, struct fs_stat *stat)
...
+}
+
+static status_t spifs_file_ioctl(filecookie *cookie, int request, void *argp)
+{
+    LTRACEF("request %d, argp %p\n", request, argp);
+
+    switch (request) {
+        case FS_IOCTL_GET_FILE_ADDR: {
+            return spifs_ioctl_get_file_addr(cookie, (void**)argp);
+        }
```
nvll: I'd toss a break here for maintenance sake, and possibly a default that can warn on a request that isn't valid.

lib/fs/fs.c
```diff
@@ -245,6 +245,13 @@ status_t fs_open_file(const char *path, filehandle **handle)
     return 0;
 }
 
+status_t fs_file_ioctl(filehandle *handle, int request, void *argp)
+{
+    LTRACEF("filehandle %p, request %d, argp, %p\n", handle, request, argp);
+
+    return handle->mount->api->file_ioctl(handle->cookie, request, argp);
```
nvll: Unsafe. You're dereferencing 4 pointers with no checks.

lib/fs/debug.c
```diff
@@ -78,6 +78,47 @@ STATIC_COMMAND_END(fs);
...
+            err = fs_open_file(argv[3].str, &handle);
+            if (err != NO_ERROR) {
+                printf("error %d opening file\n", err);
+                return err;
+            }
+
+            void *file_addr;
+            err = fs_file_ioctl(handle, request, &file_addr);
+            if (err != NO_ERROR) {
+                fs_close_file(handle);
```
nvll: is fs_close_file() safe to call on anything return by fs_file_ioctl?

lib/fs/debug.c
gkalsi: ack.

lib/fs/debug.c
gkalsi: I think this is what happens when only the spacing is changed.

lib/fs/debug.c
gkalsi: we call `fs_file_open` above and test that it succeeded so it should be safe to call close on the file handle

gkalsi:
merged

PR #84, opened 2015-12-11
Pr/colourlcd

gkalsi:
+ Added support for the 1.33 inch sharp colour lcd
+ The sharp memory lcds appear to have a pretty consistent command set so I refactored the driver to use as many common codepaths as possible
+ The LCD that's attached to the device is configured as a compile time flag


target/dartuinoP0/include/target/display/LS013B7DH06.h
```diff
@@ -0,0 +1,40 @@
...
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+// 1.33 Inch 3-Bit RGB Sharp Color LCD
+
+#ifndef TARGET_DARTUINOP0_INCLUDE_TARGET_DISPLAY_LS013B7DH06
```
travisg: BTW feel free to use #pragma once so you dont have to type in monstrosities like this

target/dartuinoP0/rules.mk
```diff
@@ -29,9 +31,30 @@ MODULE_SRCS += \
     $(LOCAL_DIR)/usb.c \
     $(LOCAL_DIR)/memory_lcd.c \
 
+ifeq ($(DISPLAY),LS013B7DH06)
```
travisg: could collapse this a bit better, since only a few lines actually change.

travisg:
LGTM, take it.

gkalsi:
merged.

PR #85, opened 2015-12-11
[nrf51xxx][arm-m][platform] Port to nrf51xxx

hollanderic:
tested the nrf51-pca10028-test project against the nordic pca10028 dev kit.

Also tested against dartuinoP0-test on the dartuino hardware
Also tested against stm32f4-discovery-test on the f4 disco

### DIFF
--- a/arch/arm/arm-m/systick/systick.c
+++ b/arch/arm/arm-m/systick/systick.c
@@ -109,7 +109,9 @@ lk_time_t current_time(void)
     do {
         t = ticks;
         delta = (volatile uint32_t)SysTick->VAL;
+#if       (__CORTEX_M >= 0x03)
         DMB;
+#endif
     } while (ticks != t);
 
     /* convert ticks to msec */
@@ -128,7 +130,9 @@ lk_bigtime_t current_time_hires(void)
     do {
         t = ticks;
         delta = (volatile uint32_t)SysTick->VAL;
+#if       (__CORTEX_M >= 0x03)
         DMB;
+#endif        
     } while (ticks != t);
 
     /* convert ticks to usec */
--- a/arch/arm/include/arch/arch_ops.h
+++ b/arch/arm/include/arch/arch_ops.h
@@ -280,29 +280,145 @@ static inline void set_current_thread(struct thread *t)
 #else // pre-armv6 || (armv6 & thumb)
 
 /* for pre-armv6 the bodies of these are too big to inline, call an assembly stub version */
-void _arch_enable_ints(void);
-void _arch_disable_ints(void);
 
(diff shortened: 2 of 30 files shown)


arch/arm/arm-m/systick/systick.c
```diff
@@ -109,7 +109,9 @@ lk_time_t current_time(void)
     do {
         t = ticks;
         delta = (volatile uint32_t)SysTick->VAL;
+#if       (__CORTEX_M >= 0x03)
         DMB;
```
travisg: probably better to ifdef out the definition of DMB in the header than do it on every use.

arch/arm/include/arch/arch_ops.h
```diff
@@ -280,29 +280,145 @@ static inline void set_current_thread(struct thread *t)
...
-static inline int atomic_add(volatile int *ptr, int val) { return _atomic_add(ptr, val); }
-static inline int atomic_and(volatile int *ptr, int val) { return _atomic_and(ptr, val); }
-static inline int atomic_or(volatile int *ptr, int val) { return _atomic_or(ptr, val); }
-static inline int atomic_swap(volatile int *ptr, int val) { return _atomic_swap(ptr, val); }
-static inline int atomic_cmpxchg(volatile int *ptr, int oldval, int newval) { return _atomic_cmpxchg(ptr, oldval, newval); }
+    __asm__ volatile("mrs %0, primask" : "=r"(state));
+    state &= 0x1;
+    return !!state;
+}
+static uint32_t _arch_critical_region_counter = 0;
```
travisg: this isn't good. This variable must be declared in one spot in a .c file somewhere, or every .c file that sees this .h will create a separate one.

platform/nrf51xxx/uart.c
```diff
@@ -0,0 +1,128 @@
...
+
+int uart_getc(int port, bool wait)
+{
+    do {
+	    if (NRF_UART0->EVENTS_RXDRDY > 0) {
+                NRF_UART0->EVENTS_RXDRDY=0;
+                return NRF_UART0->RXD;
+        }
+    } while (wait);
+    return 0;
```
travisg: return a -1 here if !wait and you dont have a new character. the reader is probably interpreting it as a long stream of \0 at the moment. Which may or may not work.

arch/arm/include/arch/arch_ops.h
travisg: Actually now that i think about it, the critical region stuff doesn't properly stack with code that doesn't use it.

The better solution here is to instead of creating this critical region stuff, just store the result of arch_ints_disabled(), then conditionally reenable it based on that.

PR #86, opened 2015-12-16
Pr/nrf51xxx

hollanderic:
-Fixed bugs in pendsv
-Changed the method for disabling irqs in atomic functions
-Using RTC0 instead of systick (systick not implemented on nRF51822)

### DIFF
--- a/app/tests/benchmarks.c
+++ b/app/tests/benchmarks.c
@@ -158,12 +158,14 @@ __NO_INLINE static void arm_bench_cset_stm(void)
 __NO_INLINE static void arm_bench_multi_issue(void)
 {
     uint32_t cycles;
+#if       (__CORTEX_M >= 0x03)
     uint32_t a = 0, b = 0, c = 0, d = 0, e = 0, f = 0, g = 0, h = 0;
-
+#endif
 #define ITER 1000000
     uint count = ITER;
     cycles = arch_cycle_count();
     while (count--) {
+#if       (__CORTEX_M >= 0x03)
         asm volatile ("");
         asm volatile ("add %0, %0, %0" : "=r" (a) : "r" (a));
         asm volatile ("add %0, %0, %0" : "=r" (b) : "r" (b));
@@ -173,6 +175,17 @@ __NO_INLINE static void arm_bench_multi_issue(void)
         asm volatile ("add %0, %0, %0" : "=r" (f) : "r" (f));
         asm volatile ("and %0, %0, %0" : "=r" (g) : "r" (g));
         asm volatile ("mov %0, %0" : "=r" (h) : "r" (h));
+#else
+        asm volatile ("");
+        asm volatile ("add r0, r0":::"r0");
+        asm volatile ("add r0, r0":::"r0");
+        asm volatile ("and r0, r0":::"r0");
+        asm volatile ("mov r0, r0":::"r0");
+        asm volatile ("orr r0, r0":::"r0");
+        asm volatile ("add r0, r0":::"r0");
+        asm volatile ("and r0, r0":::"r0");
+        asm volatile ("mov r0, r0":::"r0");
+#endif
     }
     cycles = arch_cycle_count() - cycles;
 
--- a/arch/arm/arm-m/arch.c
+++ b/arch/arm/arm-m/arch.c
@@ -39,10 +39,11 @@ unsigned int arm_cm_irq_pri_mask;
 
 void arch_early_init(void)
 {
-    uint i;
 
     arch_disable_ints();
 
+#if     (__CORTEX_M >= 0x03) || (CORTEX_SC >= 300)
+    uint i;
     /* set the vector table base */
     SCB->VTOR = (uint32_t)&vectab;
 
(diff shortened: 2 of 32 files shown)


app/tests/benchmarks.c
```diff
@@ -158,12 +158,14 @@ __NO_INLINE static void arm_bench_cset_stm(void)
 __NO_INLINE static void arm_bench_multi_issue(void)
 {
     uint32_t cycles;
+#if       (__CORTEX_M >= 0x03)
```
travisg: may as well just #ifdef out this entire function, since there's no point trying to benchmark multi-issue on a non multi-issue cpu.

arch/arm/arm-m/arch.c
```diff
@@ -72,11 +73,11 @@ void arch_early_init(void)
 
     /* set priority grouping to 0 */
     NVIC_SetPriorityGrouping(0);
-
     /* enable certain faults */
-    SCB->SHCSR |= (SCB_SHCSR_USGFAULTENA_Msk | SCB_SHCSR_BUSFAULTENA_Msk | SCB_SHCSR_MEMFAULTENA_Msk);
 
+    SCB->SHCSR |= (SCB_SHCSR_USGFAULTENA_Msk | SCB_SHCSR_BUSFAULTENA_Msk | SCB_SHCSR_MEMFAULTENA_Msk);
```
travisg: is this a whitespace difference?

arch/arm/arm-m/arch.c
```diff
@@ -142,13 +147,11 @@ void arm_cm_irq_entry(void)
 
 void arm_cm_irq_exit(bool reschedule)
 {
-    target_set_debug_led(1, false);
```
travisg: put this back

arch/arm/arm-m/exceptions.c
```diff
@@ -40,14 +41,13 @@ static void dump_frame(const struct arm_cm_exception_frame *frame)
     printf("\tlr  0x%08x pc  0x%08x psr 0x%08x\n",
            frame->lr, frame->pc, frame->psr);
 }
-
```
travisg: try not to make too many whitespace changes like this

arch/arm/arm-m/thread.c
```diff
@@ -87,17 +87,17 @@ void arch_thread_initialize(struct thread *t)
 
 volatile struct arm_cm_exception_frame_long *preempt_frame;
 
+static void pendsv(struct arm_cm_exception_frame_long *frame) asm("pendsv");
```
travisg: hmm, what does the asm("pendsv") do? is that some new gcc sorcery?

arch/arm/arm-m/thread.c
```diff
@@ -87,17 +87,17 @@ void arch_thread_initialize(struct thread *t)
...
+
 static void pendsv(struct arm_cm_exception_frame_long *frame)
 {
     arch_disable_ints();
 
-    LTRACEF("preempting thread %p (%s)\n", _current_thread, _current_thread->name);
```
travisg: put these back.

arch/arm/arm-m/thread.c
```diff
@@ -118,10 +121,48 @@ __NAKED void _pendsv(void)
         "pop	{ r4-r11, lr };"
         "bx		lr;"
         :: "i" (pendsv)
+        );
+    __UNREACHABLE;
+#else
+    struct arm_cm_exception_frame_long *frame;
+
+    __asm__ volatile(
```
travisg: I'd be very careful about this code. Theoretically you cannot emit any C code in a _NAKED function since there was no stack frame set up. I think this *might* be okay because the state is saved in asm, but I'm not so sure it's okay. That's the reason I had the convoluted branch to another function that did the body of this in the M3 version.

arch/arm/arm-m/thread.c
```diff
@@ -150,6 +204,26 @@ __NAKED static void _half_save_and_svc(vaddr_t *fromsp, vaddr_t tosp)
...
+        "mov    r2, r8;"
+        "mov    r3, r9;"
+        "push   { r2-r3 };"
+        "push   { r4-r7 };"
+
+        "mov    r3, sp;"
+        "str	r3, [r0];"
+
+        /* make sure we load the destination sp here before we reenable interrupts */
+        "mov	sp, r1;"    /* Seems like a double tap here, sp gets loaded with same */
```
travisg: the comment doesn't make sense, it only loads SP once.

arch/arm/include/arch/arch_ops.h
```diff
@@ -280,29 +280,146 @@ static inline void set_current_thread(struct thread *t)
...
+    __asm__ volatile("cpsid i");
+    CF;
+}
+
+static inline bool arch_ints_disabled(void)
+{
+    unsigned int state;
 
-static inline uint32_t arch_cycle_count(void) { return _arch_cycle_count(); }
+    __asm__ volatile("mrs %0, primask" : "=r"(state));
```
travisg: all of this fouls up the support for pre v6 cpus, since this is replacing the 'else' clause of all of the versions with essentially m0 support.

travisg:
The platform/target stuff is fine, though we should split that into a separate commit so there is a commit for m0 support and a commit for platform/target separately.

The m0 stuff is mostly okay, though it I think breaks a good chunk of v6 support, which is pretty distinctly different from v6m (confusing as hell). But, I think that may just be easier to fix in a branch instead of via this interface.

PR #89, opened 2016-01-12
Pr/cortexm0

hollanderic:
support for cortex-m0 architecture
platform support for Nordic Semiconductor nrf51xxx
target support for Nordic development kits pca10028 and pca10000


travisg:
Looks good! I'll take it in a bit.

PR #93, opened 2016-01-18
[arch][arm] changes to support cortex-m0plus architecture

nqbit:
Add support for cortex-m0+ architecture. DebugMonitor_IRQ does not seem to be supported on all devices.


travisg:
looks good, taken.

PR #94, opened 2016-01-28
Dartuino DMA Bug

gkalsi:


gkalsi:
Here's a [diff that omits the cosmetic changes](https://github.com/littlekernel/lk/compare/20513a9901e4b05d366e42a8844166bb6c4648e0...0377b90b566d7731ee4772cc08e2a2a0ca7e110e)

platform/stm32f7xx/qspi.c
```diff
@@ -75,14 +81,36 @@ static uint32_t get_address_size(uint32_t address);
...
+static status_t dma_disable(DMA_Stream_TypeDef *dma)
+{
+    // Unset the DMA Enable bit.
+    dma->CR &= ~DMA_SxCR_EN;
+
+    uint attempts = 0;
+
+    while (dma->CR & DMA_SxCR_EN) {
+
+        thread_sleep(1);
```
travisg: What is this needed for? If it's legit, put a comment as to why it's needed.

travisg:
looks pretty legit.

platform/stm32f7xx/qspi.c
gkalsi: Will add a comment.

Basically, unsetting the DMA Enable bit isn't good enough to disable the DMA controller since xfers might still be in progress. The solution is to unset the enable bit, then wait for it to actually become disabled.

travisg: That's pretty grody, is there a better way to spin on a in-progress bit and wait for that to clear? Is dma_disable used in the common path? If so, this thread sleep will definitely slow things down.

travisg:
LGTM

gkalsi:
Merged.

PR #95, opened 2016-01-28
Pr/systick priority

gkalsi:
Remove calls to `NVIC_SetPriority` inside `arm-m/CMSIS`

NB: I created a .patch file just so that we have a paper trail of any changes we made to code that we don't own. LMK if this is a path we want to go down @travisg


travisg:
LGTM

gkalsi:
Merged.

PR #101, opened 2016-02-04
Initial stm32f0xx support

konkers:


travisg:
Looks good to me. Do you want me to roll it into mainline?

konkers:
I noticed last night that the standard peripheral driver does not
include USB support.  I think I'm going to move over to the HAL from
the stm32cube package (which does include USB.)  Your choice if you'd
like to merge this pull request or wait till I do that.

Cheers,
   Erik

travisg:
Yeah, for some reason USB is usually a separate thing. Sometimes it's not
that useful either, since the USB drivers will tend to be a bit more
involved than the average little peripheral and the ST driver library
expects a bit more control over the system.

konkers:
Then, it's probably worth merging this.... boy am I not looking forward to
writing another USB device driver/stack.....

travisg:
There is a USB device stack that you can wire into, and I've written a
driver for the STM32F7 usb peripheral, though it's unlikely to be similar
to the one you have on the f0. But, in general the device stack in dev/usb
handles most of the low to mid level packet parsing for you, you just have
to provide a way to queue transfers on all the endpoints and call back into
the stack when stuff happens.

But yeah, it's not a lot of fun.

konkers:
cool, I'll look into it.

travisg:
Okay, pushed into mainline. Closing this one out.

PR #102, opened 2016-02-05
[fs][spifs][qspi] Reduce stack usage by 12% of the total default stac…

gkalsi:
…k size for some critical call paths through the spifs filesystem layer.


travisg:
LGTM

gkalsi:
Merged.

PR #103, opened 2016-02-06
Stm32f0xx CAN support.

konkers:


konkers:
Doh, doesn't compile.  New one incoming.

PR #107, opened 2016-02-09
Shrink string usage in macros.

david-dee:
The various EXPECT_EQ, etc., macros used concatenation to build
their strings.  This creates a full error string for each
EXPECT_* test in a unittest.  Instead, use string substitution
(%s) so that the common portions of the strings can be
combined.


travisg:
LGTM, going to edit the comment to prefix with [lib][unittest] to be consistent with the rest of the commits and push it.

PR #108, opened 2016-02-11
[dev][accelerometer] Add an API to query whether or not an accelerome…

gkalsi:
…ter is present


travisg:
Hmm, it's doable, though not a good long term solution. If we went with a structured driver api i'd hope that dynamic discovery would solve it better.

I was going to suggest not making this a public api and just putting it in app/sod that target then overrides, but then I guess it does bring up the problem that target/dartuino is not in the same tree and wouldn't compile because it wouldn't be able to find the app/sod includes...

So, this may be an okay solution for now.

gkalsi:
Dropping this for now – I think there's a cleaner way to do this.

PR #109, opened 2016-02-18
Pr/portgroup add rem

gkalsi:
Add an API to dynamically add/remove ports from a port group.

### DIFF
--- a/app/tests/port_tests.c
+++ b/app/tests/port_tests.c
@@ -583,6 +583,75 @@ int group_basic(void)
     return 0;
 }
 
+int group_dynamic(void)
+{
+    status_t st;
+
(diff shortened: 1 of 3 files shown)


gkalsi:
Updated as per @cpu-chromium's comments. 

It turns out that adding a port to a port group twice is already impossible because once a read port is bound to a port group, it cannot be added to a port group again.

Note that this does not preclude opening two different ports with the same name and adding them to the same port group.

gkalsi:
@cpu-chromium Do my most recent changes look okay? Clear to merge?

cpizano:
Go ahead.

gkalsi:
Merged.

PR #110, opened 2016-02-18
pr/external

travisg:
Move code in the tree that came from external sources into a top level directory called external/. Maintains a parallell include path for the build system, so almost nothing actually changes, just moving of files around.

Take a look at see if anything offends you.


swetland:
So the only thing I'm not sure about here is that the actual external "packages" are hidden down in the bowels of the directories under external/...

The alternative is more verbose, but would make the various external sources more immediately clear, with something like:
external/arm-cmsis/arch/arm/arm-m/CMSIS/...
external/sunsoft-libm/lib/libm/...
external/lwip/lib/lwip/...
external/lpcopen-lpc15xx/platform/lpc15xx/lpcopen/...

and in this model one could mandate an external/<thing>/META or somesuch which would indicate the source of the external project, the license, etc, which would allow for easy automated processing related to license compliance, etc.  It also allows for (if someday desirable) more finegrained inclusion or exclusion of various external sources by virtue of their path/packagename.

and, maybe better done in a second pass refactor, I'd tend to treat the vendor driver libs more as libraries:
external/lpcopen-lpc15xx/lib/lpcopen/...

travisg:
Hmm, that's true, though I don't think it's worth changing in this commit. Just getting it moved over is I think the general idea here.

Looking at it again there is probably a little bug in the way external is added to the LKINC dir, will have to test that and patch it before submitting.

PR #111, opened 2016-02-18
[init] removed test code from target init

hollanderic:
Removed test code that was accidentally left in init for the pca10000 development kit target.


travisg:
LGTM

PR #114, opened 2016-02-29
Add a Truncate Primitive to FS

gkalsi:
Added a truncate primitive to the FS API layer as well as the two FS implementations.

### DIFF
--- a/lib/fs/memfs/memfs.c
+++ b/lib/fs/memfs/memfs.c
@@ -253,6 +253,42 @@ static ssize_t memfs_read(filecookie *fcookie, void *buf, off_t off, size_t len)
     return len;
 }
 
+static status_t memfs_truncate(filecookie *fcookie, uint64_t len)
+{
+    LTRACEF("filecookie %p, len %llu\n", fcookie, len);
+
+    status_t rc = NO_ERROR;
+
+    memfs_file_t *file = (memfs_file_t *)fcookie;
+
+    mutex_acquire(&file->fs->lock);
+
+    // Can't use truncate to grow a file.
+    if (len > file->len)
+        return ERR_INVALID_ARGS;
+
+    void *ptr = realloc(file->ptr, len);
+    if (unlikely(ptr == NULL)) {
+        rc = ERR_NO_MEMORY;
+        goto finish;
+    }
+
+    // Fast-path: It's very likely that realloc won't move the allocated memory
+    // block since we're shrinking an existing file. If the block hasn't been
+    // moved, we can terminate early.
+    if (ptr == file->ptr) {
+        rc = NO_ERROR;
+        goto finish;
+    }
+
+    // If we get a different pointer back, we need to move the file.
+    file->ptr = memmove(ptr, file->ptr, len);
+
+finish:
+    mutex_release(&file->fs->lock);
+    return rc;
+}
+
 static ssize_t memfs_write(filecookie *fcookie, const void *buf, off_t off, size_t len)
 {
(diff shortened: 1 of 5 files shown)


lib/fs/memfs/memfs.c
```diff
@@ -253,6 +253,42 @@ static ssize_t memfs_read(filecookie *fcookie, void *buf, off_t off, size_t len)
...
+
+    memfs_file_t *file = (memfs_file_t *)fcookie;
+
+    mutex_acquire(&file->fs->lock);
+
+    // Can't use truncate to grow a file.
+    if (len > file->len)
+        return ERR_INVALID_ARGS;
+
+    void *ptr = realloc(file->ptr, len);
```
cpu-chromium: I think we need to guard against len == 0,  nice realloc( , 0) is not well defined? 

If so maybe protect against that at the top so we don't introduce differences on the implementations?

lib/fs/memfs/memfs.c
```diff
@@ -253,6 +253,42 @@ static ssize_t memfs_read(filecookie *fcookie, void *buf, off_t off, size_t len)
...
+
+    status_t rc = NO_ERROR;
+
+    memfs_file_t *file = (memfs_file_t *)fcookie;
+
+    mutex_acquire(&file->fs->lock);
+
+    // Can't use truncate to grow a file.
+    if (len > file->len)
+        return ERR_INVALID_ARGS;
```
travisg: need to unlock the mutex here.

lib/fs/memfs/memfs.c
travisg: Pretty sure all the reallocs are fine and ,0 will produce a null... but then the line before will fail.

lib/fs/memfs/memfs.c
```diff
@@ -253,6 +253,42 @@ static ssize_t memfs_read(filecookie *fcookie, void *buf, off_t off, size_t len)
...
+    // Fast-path: It's very likely that realloc won't move the allocated memory
+    // block since we're shrinking an existing file. If the block hasn't been
+    // moved, we can terminate early.
+    if (ptr == file->ptr) {
+        rc = NO_ERROR;
+        goto finish;
+    }
+
+    // If we get a different pointer back, we need to move the file.
+    file->ptr = memmove(ptr, file->ptr, len);
```
travisg: I think you forgot to set file->len.

gkalsi:
Any more thoughts on this one @cpu-chromium or @travisg?

lib/fs/memfs/memfs.c
travisg: That's not how realloc works. It actually does the memmove for you and frees the old block (if that's how it chooses to do it). So, if you get a valid pointer back from realloc it's okay to use. The only real edge case is if it returns a null when you didn't expect it and thus you have to restore the old pointer.

lib/fs/memfs/memfs.c
```diff
@@ -253,6 +253,40 @@ static ssize_t memfs_read(filecookie *fcookie, void *buf, off_t off, size_t len)
...
+
+    mutex_acquire(&file->fs->lock);
+
+    // Can't use truncate to grow a file.
+    if (len > file->len) {
+        rc = ERR_INVALID_ARGS;
+        goto finish;
+    }
+
+    file->len = len;
```
travisg: Set this with file->ptr in case the realloc failed.

gkalsi:
Merged.

PR #117, opened 2016-03-04
[arch][x86]fix compile failure in 32bit interrupt frame structure

bzhu5:
When PAGE_FAULT_DEBUG_INFO flag is defined, the compile will fail.
This fix is to unify the member names of interrupt frame structure
x86_iframe_t. This patch also enables pagefault debugging for 32bit


travisg:
looks good, commited on master

PR #118, opened 2016-03-08
[arch][x86] load initial stack pointer in boot entry

bzhu5:
Initially, stack pointer (esp) is undefined in multiboot protocol,
to avoid potential memory corruption, set it with defined value.


travisg:
LGTM. merged into master

PR #119, opened 2016-03-10
Dartuino Bootloader

gkalsi:
### What's included ###
  + USB Boot
    - Query Build Signature
    - Boot Device 
    - Flash System image
    - Host side python tool for driving bootloader

  + FS Boot
    - Load and install a Bootimage from Spiflash

  + Booting
    - Chain Load into another image

### What's missing ###
  + USB Boot
    - USB no-op command that stops USB boot without taking any action
    - Any sort of image verification via USB (bootimage or the like)

  + FS Boot
    - Putting images on the SPIFlash
    - Signalling to the SPIFlash that an image is ready to be installed, right now if there's an image called "system.img" that is a valid bootimage, we'll try to install it

  + Booting
    - Limited shutdown of the bootloader before jumping to the main image it's on my todo list to perform a more robust shutdown.


nvll:
Roger.

travisg:
This looks pretty good to me, I'd advocate getting it in and iterating on that.

nvll:
+2

gkalsi:
Merged!

PR #122, opened 2016-03-22
Add API used by Dartino to get arena list

ErikCorryGoogle:
Please take a look. Something like this is needed for newer Dartino's that need to know where the Dart heap is going to be allocated.

### DIFF
--- a/include/kernel/novm.h
+++ b/include/kernel/novm.h
@@ -45,4 +45,11 @@ void *novm_alloc_unaligned(size_t *size_return);
 
 void novm_add_arena(const char *name, uintptr_t arena_start, uintptr_t arena_size);
 
+struct page_range {
+    void* address;
+    size_t size;
+};
+
+int novm_get_arenas(struct page_range* ranges, int number_of_ranges);
+
 #endif
--- a/include/kernel/vm.h
+++ b/include/kernel/vm.h
@@ -179,6 +179,9 @@ size_t pmm_free_kpages(void *ptr, uint count);
 /* physical to virtual */
 void *paddr_to_kvaddr(paddr_t pa);
 
+/* a hint as to which virtual addresses will be returned by pmm_alloc_kpages */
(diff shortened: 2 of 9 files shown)


include/kernel/novm.h
```diff
@@ -45,4 +45,11 @@ void *novm_alloc_unaligned(size_t *size_return);
 
 void novm_add_arena(const char *name, uintptr_t arena_start, uintptr_t arena_size);
 
+struct page_range {
+  void* address;
```
travisg: 4 space indent

include/kernel/vm.h
```diff
@@ -179,6 +179,10 @@ size_t pmm_free_kpages(void *ptr, uint count);
 /* physical to virtual */
 void *paddr_to_kvaddr(paddr_t pa);
 
+/* a hint as to which virtual addresses will be returned by pmm_alloc_kpages */
+void *first_mapping_address(void);
```
travisg: Would rather name it to be semi-consistent something like kvaddr_get_range() that returns a void *, size_t. The implementation detail of how it works should be hidden by the implementation and will probably be made smarter at some point.

travisg:
Looks pretty good. Please collapse these changes, add a [heap] prefix and push it in.

kernel/novm/novm.c
```diff
@@ -64,6 +64,17 @@ extern int _end_of_ram;
 #endif
 struct novm_arena arena[NOVM_MAX_ARENAS];
 
+int novm_get_arenas(struct page_range* ranges, int number_of_ranges)
+{
+    int ranges_found = 0;
+    for (int i = 0; i < number_of_ranges && i < NOVM_MAX_ARENAS; i++) {
+        if (arena[i].pages > 0) ranges_found = i + 1;
```
sgjesse: assert `ranges_found == i` in this if?

Right now there are no way of removing arena, so it will definitely hold.

PR #125, opened 2016-03-29
[usb][cdcserial] Single Channel CDC Serial.

gkalsi:
Here's a first pass at a CDC serial driver in LK.
Right now it consists of a single LK app that echoes characters back over the virtual COM port. 
Only a single channel is supported -- multi channel coming soon.

### DIFF
--- a/app/cdcserialtest/cdcserialtest.c
+++ b/app/cdcserialtest/cdcserialtest.c
@@ -22,7 +22,37 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-void append_usb_interfaces(void)
+#include <app.h>
+#include <err.h>
+#include <dev/usb/class/cdcserial.h>
+#include <kernel/thread.h>
+
+uint8_t rxbuf[64];
+
+static void cdctest_init(const struct app_descriptor *app)
 {
-	// no-op
+    cdcserial_start();
 }
+
+// Read bytes from CDC Serial and write them back to the stream.
+static void cdctest_entry(const struct app_descriptor *app, void *args)
+{
+    while (true) {
+        ssize_t bytes = cdcserial_read(64, rxbuf);
+        if (bytes == ERR_NOT_READY) {
+            // USB is not ready yet.
+            thread_sleep(100);
+            continue;
+        } else if (bytes < 0) {
+            printf("Error reading bytes from CDC Serial: %ld\n", bytes);
+            break;
+        }
+
+        cdcserial_write(bytes, rxbuf);
+    }
+}
+
+APP_START(usbtest)
+ .init = cdctest_init,
+ .entry = cdctest_entry,
+APP_END
--- a/app/cdcserialtest/cdcserialtest.c
+++ b/app/cdcserialtest/cdcserialtest.c
@@ -27,18 +27,17 @@
 #include <dev/usb/class/cdcserial.h>
 #include <kernel/thread.h>
 
-uint8_t rxbuf[64];
+static uint8_t rxbuf[64];
 
 static void cdctest_init(const struct app_descriptor *app)
 {
-    cdcserial_start();
 }
 
 // Read bytes from CDC Serial and write them back to the stream.
 static void cdctest_entry(const struct app_descriptor *app, void *args)
 {
     while (true) {
-        ssize_t bytes = cdcserial_read(64, rxbuf);
+        ssize_t bytes = cdcserial_read(sizeof(rxbuf), rxbuf);
         if (bytes == ERR_NOT_READY) {
             // USB is not ready yet.
             thread_sleep(100);
--- a/dev/usb/class/cdcserial/cdcserial.c
+++ b/dev/usb/class/cdcserial/cdcserial.c
@@ -1,0 +1,331 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ * Author: gkalsi@google.com (Gurjant Kalsi)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files
+ * (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge,
+ * publish, distribute, sublicense, and/or sell copies of the Software,
+ * and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include <dev/usb/class/cdcserial.h>
+
+#include <assert.h>
+#include <dev/udc.h>
+#include <dev/usb.h>
+#include <dev/usbc.h>
+#include <dev/usbc.h>
+#include <err.h>
+#include <kernel/event.h>
+#include <sys/types.h>
+#include <trace.h>
+
+#define LOCAL_TRACE 0
+
+#define W(w) (w & 0xff), (w >> 8)
+#define W3(w) (w & 0xff), ((w >> 8) & 0xff), ((w >> 16) & 0xff)
+
+#define CDC_REQ_SEND_CMD               0x00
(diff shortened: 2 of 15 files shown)


app/cdcserialtest/cdcserialtest.c
```diff
@@ -0,0 +1,58 @@
...
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include <app.h>
+#include <err.h>
+#include <dev/usb/class/cdcserial.h>
+#include <kernel/thread.h>
+
+uint8_t rxbuf[64];
```
travisg: static or put it in the routine on the stack (64 bytes should be okay for stack)

dev/usb/class/cdcserial/cdcserial.c
```diff
@@ -0,0 +1,331 @@
...
+};
+
+static status_t usb_register_cb(
+    void *cookie,
+    usb_callback_op_t op,
+    const union usb_callback_args *args
+)
+{
+    if (op == USB_CB_ONLINE) {
+        usbc_setup_endpoint(1, USB_IN, 0x40, USB_BULK);
```
travisg: Maybe future change, but should allow configuring the endpoints this will occupy. IIRC there's a routine that lets you patch endpoints in a descriptor, and then it's a matter of squirreling the endpoint numbers away in a data structure.

dev/usb/class/cdcserial/cdcserial.c
```diff
@@ -0,0 +1,331 @@
...
+                break;
+        }
+    }
+
+    return NO_ERROR;
+}
+
+status_t cdcserial_start(void)
+{
+    usb_setup(&config, &usb_setup_cb);
```
travisg: The target is really supposed to do the usb_setup, then call into <class>_start which adds their interfaces. This allows multiple classes to be present at a time.

That'll probably mess up your usb_setup_cb, since each class may register their own. Will probably need a list of callbacks, or feed them through the regular usb callback that was already there.

dev/usb/class/cdcserial/cdcserial.c
```diff
@@ -0,0 +1,331 @@
...
+        usbc_setup_endpoint(1, USB_IN, 0x40, USB_BULK);
+        usbc_setup_endpoint(1, USB_OUT, 0x40, USB_BULK);
+        usbc_setup_endpoint(2, USB_IN, 0x40, USB_INTR);
+
+        usb_online = true;
+    }
+    return NO_ERROR;
+}
+
+static status_t usb_setup_cb(
```
travisg: as pointed out below, you dont really need a separate callback here, this can be funneled through the regular CB as another opcode. Returns from all the register callbacks gets a little more complicated in that case, but if you call all of the registered CB's and then check if any of them handled the setup message it should work.

app/cdcserialtest/cdcserialtest.c
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
```
nvll: Isn't this the same copyright stuff we talked about before?

app/cdcserialtest/cdcserialtest.c
```diff
@@ -0,0 +1,58 @@
...
+static void cdctest_init(const struct app_descriptor *app)
+{
+    cdcserial_start();
+}
+
+// Read bytes from CDC Serial and write them back to the stream.
+static void cdctest_entry(const struct app_descriptor *app, void *args)
+{
+    while (true) {
+        ssize_t bytes = cdcserial_read(64, rxbuf);
```
nvll: sizeof(rxbuf). 64 is dangerous if rxbuf ever changes.

app/cdcserialtest/cdcserialtest.c
```diff
@@ -0,0 +1,58 @@
...
+        if (bytes == ERR_NOT_READY) {
+            // USB is not ready yet.
+            thread_sleep(100);
+            continue;
```
nvll: Is 100 ms arbitrary?

app/cdcserialtest/cdcserialtest.c
```diff
@@ -0,0 +1,58 @@
...
+        } else if (bytes < 0) {
+            printf("Error reading bytes from CDC Serial: %ld\n", bytes);
```
nvll: If I'm reading this right we never recover from a read error? Is it reasonably recoverable?

dev/usb/class/cdcserial/cdcserial.c
```diff
@@ -0,0 +1,331 @@
...
+
+/* high/low speed device qualifier */
+static const uint8_t devqual_descr[] = {
+    0x0a,             /* descriptor length */
+    DEVICE_QUALIFIER, /* Device Qualifier type */
+    W(0x0200),        /* USB version */
+    0x00,             /* class */
+    0x00,             /* subclass */
+    0x00,             /* protocol */
+    64,               /* max packet size for endpoint 0 */
```
nvll: Same size as rxbuf?

dev/usb/class/cdcserial/cdcserial.c
```diff
@@ -0,0 +1,331 @@
...
+    0x06,   /* bDescriptorSubtype: Union func desc */
+    0x00,   /* bMasterInterface: Communication class interface */
+    0x01,   /* bSlaveInterface0: Data Class Interface */
+
+    /* endpoint 1 IN */
+    0x07,           /* length */
+    ENDPOINT,       /* type */
+    0x82,           /* address: 1 IN */
+    0x03,           /* type: bulk */
+    W(8),           /* max packet size: 64 */
```
nvll: Same as above.

dev/usb/class/cdcserial/cdcserial.c
```diff
@@ -0,0 +1,331 @@
...
+    0x07,      /* bLength: Endpoint Descriptor size */
+    ENDPOINT,  /* bDescriptorType: Endpoint */
+    0x81,      /* bEndpointAddress */
+    0x02,      /* bmAttributes: Bulk */
+    0x40,      /* wMaxPacketSize: */
+    0x00,
+    0x00       /* bInterval */
+};
+
+static const uchar langid[] = { 0x04, 0x03, 0x09, 0x04 };
```
nvll: What do these numbers mean?

dev/usb/class/cdcserial/cdcserial.c
nvll: Should 0x40 be a define?

dev/usb/class/cdcserial/cdcserial.c
travisg: Actually no. It's the transfer size of a single usb packet. 0x40 is the maximum size for a control transfer. Technically this should come from the configuration of the device controller, but no real device supports less than the max.

dev/usb/class/cdcserial/cdcserial.c
travisg: I forget actually. It's something to do with the encoding of any string descriptors, and this is the generic one. But as I pointed out this particular part is part of the overall usb config and actually doesn't belong here anyway, it's up to the target/project to define the overall device descriptor. This code should just append some interface descriptor to the list.

app/cdcserialtest/cdcserialtest.c
gkalsi: I think the consensus was that assigning copyright to Google was okay, this one is assigned to Google.

app/cdcserialtest/cdcserialtest.c
gkalsi: Pretty much. I don't expect that we'll ever use `app/cdcserialtest` ... It's more of a demo to show that we can get a bi-di stream of bytes

app/cdcserialtest/cdcserialtest.c
gkalsi: Ditto to my comment above.

dev/usb/class/cdcserial/cdcserial.c
```diff
@@ -0,0 +1,300 @@
...
+{
+    LTRACEF("len = %d, buf = %p\n", len, buf);
+
+    DEBUG_ASSERT(buf);
+
+    if (!usb_online) {
+        return ERR_NOT_READY;
+    }
+
+    usbc_transfer_t transfer = {
```
travisg: as i pointed out in other change, i think you want this to be static because the usbc code will hold onto a pointer to this transfer and pass it back in the callback.

gkalsi: However the code in `cdcserial_write` won't make progress until that callback has signaled for it to continue.
Doesn't that mean that we can guarantee that the scope of `transfer` exceeds the duration for which the usb stack will hold on to it?

travisg: Ah okay. Yeah, it wont stick around once it calls back, so you should be okay.

travisg:
This seems pretty okay to me. I dont see any major issues with it.

gkalsi:
Merged.

PR #126, opened 2016-03-29
[stm32f7][qspi][debug] Add tracing to QSPI driver and STM32f7xx QPSI …

gkalsi:
…library. Reset QSPI Controller during init.


travisg:
LGTM. unless you think you're gonna break something, don't bother PRing this like this, just shove it in.

gkalsi:
ack, old habits die hard I guess

PR #127, opened 2016-04-13
High Level Debugger - Bidirectional Serial Pipe Over USB

gkalsi:
This should be a good transport to carry the Dartino debugger over.


travisg:
Should put a license on the files.

lib/ndebug/include/lib/ndebug/shared_structs.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+struct __attribute__((__packed__)) ndebug_ctrl_packet_t {
```
travisg: use compiler.h and __PACKED

lib/ndebug/include/lib/ndebug/ndebug.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <stdint.h>
+#include <sys/types.h>
+
+#define NDEBUG_MAX_PACKET_SIZE (64)
```
travisg: make sure you cover all header files with __BEGIN_CDECLS and __END_CDELCS

lib/ndebug/include/lib/ndebug/shared_structs.h
```diff
@@ -0,0 +1,18 @@
...
+    uint32_t magic;
```
travisg: probably need stdint.h for the uint32_ts

lib/ndebug/ndebug.c
```diff
@@ -0,0 +1,190 @@
...
+    usbc_transfer_t transfer = {
+        .callback = &usb_recv_cplt_cb,
+        .result = 0,
+        .buf = buf,
+        .buflen = n,
+        .bufpos = 0,
+        .extra = (void *)ch,
+    };
+
+    usbc_queue_rx(CH_TO_ADDR(ch), &transfer);
```
travisg: pretty sure the queue_rx holds onto the transfer pointer, and the one you created on the stack is going out of scope. Might need to be static or global.

lib/ndebug/ndebug.c
```diff
@@ -0,0 +1,190 @@
...
+    usbc_transfer_t transfer = {
+        .callback = &usb_xmit_cplt_cb,
+        .result = 0,
+        .buf = buf,
+        .buflen = n,
+        .bufpos = 0,
+        .extra = (void *)ch,
+    };
+
+    usbc_queue_tx(CH_TO_ADDR(ch), &transfer);
```
travisg: same as before with the transfer going out of scope.

lib/ndebug/ndebug.c
```diff
@@ -0,0 +1,190 @@
...
+static status_t usb_xmit_cplt_cb(ep_t endpoint, usbc_transfer_t *t)
+{
+    uint32_t channel = (uint32_t)t->extra;
+    CHECK_CHANNEL(channel);
+
+    event_signal(&tx_event[channel], false);
+    return 0;
+}
+
+static status_t usb_recv_cplt_cb(ep_t endpoint, usbc_transfer_t *t)
```
travisg: should probably check the result of the transfer and make sure it actually received something. IIRC the len field is adjusted with the amount of data actually received.

lib/ndebug/user.c
```diff
@@ -0,0 +1,176 @@
...
+
+    if (!connected) {
+        // User must call ndebugusr_await_host and establish a connection
+        // before continuing.
+        return ERR_CHANNEL_CLOSED;
+    }
+
+    // TODO(gkalsi): Packetize the packet that's too big and send the data
+    //               anyway.
+    if (n > NDEBUG_USR_MAX_PACKET_SIZE)
```
travisg: I think you want to test for this + sizeof(header)

tools/Makefile
```diff
@@ -13,5 +13,11 @@ MKIMAGE_INCS := -I../external/lib/mincrypt/include -I../lib/bootimage/include
 mkimage: $(MKIMAGE_SRCS) $(MKIMAGE_DEPS)
 	gcc -Wall -g -o $@ $(MKIMAGE_INCS) $(MKIMAGE_SRCS)
 
+NDEBUG_DEPS :=
+NDEBUG_SRCS := ndebug/ndebug.cc ndebug/usbionode.cc ndebug/tcpionode.cc
+NDEBUG_INCS := -I./ndebug/include -I../lib/ndebug/include
+ndbg: $(NDEBUG_SRCS) $(NDEBUG_DEPS)
+	g++ -std=c++11 -Wall -g -o $@ $(NDEBUG_INCS) $(NDEBUG_SRCS) -L/usr/lib/x86_64-linux-gnu -lusb-1.0
```
travisg: do you actually need the -L/usr/lib/... path? seems like that would be implicit, and it keeps it from building on mac.

lib/ndebug/include/lib/ndebug/shared_structs.h
gkalsi: Will that work when we're building the host side tools since this file is shared between the host tools and the LK build?

travisg: ah probably not.

lib/ndebug/user.c
gkalsi: I think this is fine because `NDEBUG_USR_MAX_PACKET_SIZE` is defined as `MAX_PACKET_SIZE - sizeof (header)`

lib/ndebug/ndebug.c
gkalsi: Done, except I'm checking the return value in `ndebug_usb_write` and `ndebug_usb_read` rather than in the callbacks themselves.

travisg:
LGTM

gkalsi:
Committed.

PR #129, opened 2016-04-20
[ndebug] Move ndebug reset handshake from user endpoint to all endpoi…

gkalsi:
…nts.

NDebug sits and waits for a reset message before becoming active. I intend to
use this handshake for the system debug protocol as well as the high level
protocol so I'm pulling the logic for it up into the core ndebug utility
functions.


travisg:
LGTM

PR #133, opened 2016-05-06
Pr/system debug

gkalsi:


lib/ndebug/system/consoleproxy.c
```diff
@@ -0,0 +1,83 @@
...
+                                 const char *str, size_t len)
+{
+    if (!ndebug_sys_connected()) return;
+    if (!str) return;
+
+    size_t written = 0;
+    while (written < len) {
+        size_t chunk_len = MIN(len, MAX_MUX_PACKET_SIZE);
+
+        ssize_t result = ndebug_write_sys(
```
travisg: Is this safe to call from interrupt context? The implementation looks like it grabs mutexes and whatnot. Probably should drop stuff printed at interrupt time.

platform/qemu-virt/rules.mk
```diff
@@ -37,7 +37,8 @@ MODULE_DEPS += \
 GLOBAL_DEFINES += \
     MEMBASE=$(MEMBASE) \
     MEMSIZE=$(MEMSIZE) \
-    PLATFORM_SUPPORTS_PANIC_SHELL=1
+    PLATFORM_SUPPORTS_PANIC_SHELL=1 \
+	CONSOLE_HAS_INPUT_BUFFER=1
```
travisg: Did you verify that this all works with the input buffer turned off again? This platform/target should probably be a good test case since it has code in both paths.

lib/ndebug/system/consoleproxy.c
```diff
@@ -35,6 +36,8 @@
 void consoleproxy_print_callback(print_callback_t *cb,
                                  const char *str, size_t len)
 {
+    if (arch_in_int_handler()) return;
```
gkalsi: @travisg To your previous point about calling `printf` from IRQ context, this should drop any data from IRQ context on the floor.

platform/qemu-virt/rules.mk
gkalsi: Yep, I tested it with the input buffer disabled, seems to work fine

travisg:
LGTM

gkalsi:
No longer needed.

PR #135, opened 2016-05-25
[platform][mediatek] support mediatek platform: mt6797

minglun-tsai:
Basic mt6797 platform S/W for lk.


swetland:
The license here is incompatible with lk -- it asserts that the software is proprietary and not usable without prior permission.  A more appropriate license (MIT, BSD, etc) that is compatible with LK's MIT license should be used (with permission of the copyright holder, of course).

minglun-tsai:
We are working with legal team members for the license issue, thanks.

PR #138, opened 2016-05-31
[spifs] Fix warning with GCC 5.3.1

sgjesse:
With GCC 5.3.1 (5-2016-q2-update from https://launchpad.net/gcc-arm-embedded)
additional warnings about unused variables cause compilation to fail.

### DIFF
--- a/lib/fs/spifs/spifs.c
+++ b/lib/fs/spifs/spifs.c
@@ -152,9 +152,11 @@ static status_t cursor_init(
 
 static uint8_t *cursor_get(cursor_t *cursor)
 {
+#if LK_DEBUGLEVEL > 1
     spifs_t *spifs = cursor->spifs;
 
     uint8_t *page_end = spifs->page + spifs->page_size;
+#endif
     DEBUG_ASSERT(cursor->data < page_end);
 
     return cursor->data;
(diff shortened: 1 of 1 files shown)


lib/fs/spifs/spifs.c
```diff
@@ -152,9 +152,11 @@ static status_t cursor_init(
 
 static uint8_t *cursor_get(cursor_t *cursor)
 {
+#if LK_DEBUGLEVEL > 1
     spifs_t *spifs = cursor->spifs;
 
     uint8_t *page_end = spifs->page + spifs->page_size;
+#endif
```
gkalsi: Should these `#if` and `#endif` blocks not also contain the `DEBUG_ASSERT` ?

gkalsi:
One comment, other than that it looks good. Thanks!

nvll:
The DEBUG_ASSERT itself is wrapped by a check for DEBUGLEVEL >= 1, so that's a bit redundant in the end. Either way.

lib/fs/spifs/spifs.c
sgjesse: I have no preference, and looking at the LK source found no precedence.

I will move it - it is probably more consistent.

sgjesse:
I moved the `#endif`.

PR #141, opened 2016-06-02
[console][io] Add an optional global input queue.

gkalsi:
In this model, getc reads from a global input buffer and the uart posts data to this buffer.

This commit also enables the global input queue for `STM32F7` and `QEMU-Virt`.

Tested on `STM32F7/Dartuino` and `QEMU-ARM A15` with both `CONSOLE_HAS_INPUT_BUFFER=1` and `CONSOLE_HAS_INPUT_BUFFER=0`


travisg:
LGTM, make sure the path with it disabled still works and we should be okay.

PR #152, opened 2016-08-15
Spelling fixes

ka7:
some trivial spelling fixes, just in comments.
this commit includes the external/ directory. ( just in case.. )


nvll:
Thanks. My gut says we shouldn't modify code in external if we can help it because it generally consists of unmodified third party code. Someone else chime in?

travisg:
Yeah, that's right. I'd rather not touch external/ code for spelling fixes for that reason. Everything else is good though.

ka7:
this was done via https://github.com/ka7/misspell_fixer + some prove-reading :)
( word-list is basically lintian.debian.org / spelling )

travisg:
I'm not going to take the external ones, and the other pull seems to have picked up the non-external merges, so closing this one. Thanks!

PR #157, opened 2016-08-23
Pull in remaining Pi3 Changes

gkalsi:


arch/arm64/start.S
```diff
@@ -94,6 +100,11 @@ FUNCTION(_start)
 .Lnot_device:
 
 .Lmapping_size_loop:
+    ldr     attr, =MMU_PTE_KERNEL_DATA_FLAGS
+    ldr     tmp, =arm_reset
```
hollanderic: might be nitpicky, but we might want to use __code_start here instead of arm_reset.

arch/arm64/asm.S
```diff
@@ -83,5 +83,60 @@ FUNCTION(arm64_el3_to_el1)
 
     eret
 
+FUNCTION(arm64_elX_to_el1)
```
hollanderic: Should we either save off x0 or use something >x3 in these routines since it is possible that arguments could be passed to the kernel in x0-x3?  Based on where these routines are likely to be used we could be mangling args passed to kernel.

travisg:
okay, going to take this one, with a little merge tweak in the linker script.

PR #158, opened 2016-09-01
[arm64] enable cycle counter

raphui:
Hello,

Cycle counter is not enable on arm64 arch, so the command line "bench" always return "took 0 cycles".

This patch enable cycle counter, let me know if it's ok or not.


arch/arm64/arch.c
```diff
@@ -63,6 +63,10 @@ void arch_early_init(void)
 
 void arch_init(void)
 {
+
+    int pmcr_mask = 1;
+
+    __asm__ volatile("msr pmcr_el0, %0":: "r"(pmcr_mask):);
```
travisg: Looks good, but need to move it into the percpu routines, so it also gets initialized on more than just the first cpu.

raphui:
Yes, you're right, I will do it and send a v2 pull request when I am back from holiday.

travisg:
I did another implementation over in magenta that works on all processors. I'll try to merge that back soon.

PR #159, opened 2016-09-01
[aarch64] Fix stack pointer misalignment

hollanderic:
sp can become misaligned from 16 byte boundry because context frame
is not a multiple of 16 bytes.


travisg:
lgtm

PR #160, opened 2016-09-05
[make] Compare whole words for LKINC check.

girtsf:
I had my external project directory named foo_lk, and this got me puzzled for a bit. Maybe this will save somebody else from the same issue.

My make-fu is not super strong, so please review with caution. (Also, hi, Travis!)


travisg:
LGTM

PR #161, opened 2016-09-06
[wait-queue] fix wake all to remove threads from tail

mahavirj:
While waking up all threads from wait-queue, order should
start from tail to maintain correct scheduling sequence.

Signed-off-by: Mahavir Jain <mahavirpj@gmail.com>

### DIFF
--- a/kernel/thread.c
+++ b/kernel/thread.c
@@ -1217,7 +1217,7 @@ int wait_queue_wake_all(wait_queue_t *wait, bool reschedule, status_t wait_queue
     }
 
     /* pop all the threads off the wait queue into the run queue */
-    while ((t = list_remove_head_type(&wait->list, thread_t, queue_node))) {
+    while ((t = list_remove_tail_type(&wait->list, thread_t, queue_node))) {
         wait->count--;
         DEBUG_ASSERT(t->state == THREAD_BLOCKED);
         t->state = THREAD_READY;


travisg:
No so sure about that, since it takes the items off the queue and then shoves them in the head of the run queue, if you pop from the tail it'll have the effect of putting the thread that's been waiting the longest deeper in the queue, and the thread that's been waiting the least longest at the head. With it popping from the head, it inverts the order and puts the oldest thread at the head, which i think is slightly more fair.

mahavirj:
Let us consider a scenario, where resource is acquired, and thread1 blocks on it at time t1. And later thread2 blocks at time t2 (t2 > t1). Now if there is call to wake all threads, thread2 will get chance to run earlier than thread1, which seems incorrect (thread2 being at head of run queue). Correct me if I am wrong here.

zeusk:
New threads get added to the tail, so thread2 will be at tail and not the head.

Refer lk/kernel/thread.c:1109

mahavirj:
Ping...

travisg:
Amazing, after all this time I looked at it again and you're totally right. I had been thinking at the time that the order of the threads in the queue was the other way around, but since new waiters get added to the tail, you absolutely want to pull from the tail first so that the one in the list the longest (the head) is moved into the scheduler last.

travisg:
Thanks for leaving the branch up, even after all this time I'm able to merge it.

PR #165, opened 2016-09-13
[kernel] fix get top thread error if NUM_PRIORITIES < 32

circl0:
if NUM_PRIORITIES < 32, result will be negative?


travisg:
looks good. I dont think it's ever been compiled with <32 priorities, though it does make sense to do so on memory constrained devices. 8 will probably suffice in that case.

PR #166, opened 2016-09-15
[make] don't call build engine multiple times.

girtsf:
If some included rule.mk defines an additional make target that the user
might want to call (e.g., "flash"), and user specifies a project name on
the command line, then previously the outer "makefile" would call
engine.mk multiple times. The new logic prevents that.

Additionally, silence extra "Nothing to be done" messages when calling
engine.mk again after removing the project name from make goals.


travisg:
looks good.

PR #169, opened 2016-09-30
[dev/gpio_i2c] add optional pullups

girtsf:
If GPIO_I2C_PULLUPS == 1, enables pullups when not actively
driving I2C lines low.


travisg:
LGTM

PR #170, opened 2016-10-07
[make] [lib/version] remove a misplaced .PHONY

girtsf:
When lib/version was included, a rule depended on .PHONY.  This would
unexpectedly invoke other rules defined as dependent on .PHONY (standard
phony-ness signal).

As a side effect, this used to generate srcfiles.txt and
include_paths.txt in build dir. Since they are useful, we might always
build those anyway, and not make them phony at all.

### DIFF
--- a/engine.mk
+++ b/engine.mk
@@ -73,7 +73,7 @@ ARCH_CPPFLAGS :=
 ARCH_ASMFLAGS :=
 
 # top level rule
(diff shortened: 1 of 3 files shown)


girtsf:
doh, forgot a $@.tmp and didn't retest with a clean build. fix coming.

travisg:
The latter one with buildid.h isn't really needed because the MAKECONFIGHEADER thing explicitly does the Right Thing and only replaces the file if it was different. It generates it every time but throws away the result of it's not different.

PR #175, opened 2016-10-31
[make] output .map file

piec:
Produce .map file at link time, useful to debug position of symbols, sections, etc


travisg:
LGTM

PR #188, opened 2017-05-19
Bug fix on SMAP/SMEP detection and float test.

James-Zzz:
Hi Travisg,
There is an bug on x86 architecture code: SMAP/SMEP detection, SMAP is bit 7 and SMEP is bit 20 while invoking CPU with subleaf 07H.
To float test case, since test case updated, test results for each thread should not similar now. Array of expected test result added to help check whether test result for each thread is correct.


James-Zzz:
separate to different pull request

PR #189, opened 2017-05-26
[arch][x86] SMEP/SMAP detection bug fix and memcpy/memset x86 architecture implementation

James-Zzz:
Hi Travisg,
This pull request contains 3 changes now:
1. Bug fix on Intel architecture SMEP/SMAP detection: there is an bug while calculating SMEP/SMAP bit in return value of CPUID instruction;
2. Float test: float test for each thread has different result, but message shows they should be similar. To address this issue, introduced an array to contain value for each thread, and present to uses as a reference.
3. Intel x86-64 architecture memcpy/memset implementation. Current libc function memcpy has defect while copy memory in this scenario: src < dst < src + count, content of source memory would be overwritten. To address this issue, we introduce x86-64 implementation of memcpy. Also we provided memset implementation on x86-64. Also memcpy_s introduced for secure concern.


lib/libc/string/arch/x86/64/memcpy.c
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * When decrement copy selected, processor copys data first, and then
+ * decreases RDI/RSI/RCX. In this case, we need to adjust RDI/RSI before
+ * copy to right address.
+ *
+ * More details about MOVS/MOVSB/MOVSW/MOVSD/MOVSQ please refer to
+ * <Intel 64 and IA-32 Architectures Software Developer's Manaual>
+ * Volume 2: Instruction Set Reference
+ */
+void *memcpy(void *dest, const void *src, size_t count)
```
bzhu5: is there a way to keep this function internally (static), so that others cannot call this by mistake?

lib/libc/string/arch/x86/64/memset.c
```diff
@@ -0,0 +1,79 @@
...
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+#include <string.h>
+#include <sys/types.h>
+
+static inline void inner_memset(char **dest, uint8_t val, size_t count)
+{
+    __asm__ __volatile__ (
+            "cld        \n\t"
```
bzhu5: save/restore previous DF flag.

lib/libc/string/arch/x86/64/memcpy.c
travisg: Is all this complexity actually worth it vs just going directly to rep movs? I'm generally under the impression that on modern x86s that's really the best way to do it short of using SSE.

travisg: For the magenta branch of this we ended up just using rep movs directly: https://fuchsia.googlesource.com/magenta/+/master/kernel/lib/libc/string/arch/x86/

James-Zzz: Yes, you are right, we have noticed that magenta has implemented x86 implementation, it uses movsb only. And using SSE is most efficient way to implement it.
So, do you have plan to update x86 implementation on lib/libc/string ? If yes, may be we can help implement it.
Thanks. :)

travisg: Perhaps, though the magenta kernel does not use any floating point, so using SSE is fairly expensive due to having to save the floating point context before using it. I did some informal benchmarks and didn't see a huge performance difference vs a plain rep movsb for the kernel at least.

moe123: @travisg from my experience the most efficient implementation that I found and used was the apple's xnu one.

PR #192, opened 2017-07-02
Add STM32F429I-DISCO1 support(default serial:USART1).

i25ffz:
Add STM32F429I-DISCO1 support(default serial:USART1).


i25ffz:
Run scripts/travis-build.sh error.

PR #196, opened 2017-07-29
Fix two multiboot for x86

xiw:
Several multiboot constants are off per the multiboot spec, leading to incorrect booting:

https://www.gnu.org/software/grub/manual/multiboot/multiboot.html


travisg:
Thanks for the commit!

PR #213, opened 2017-12-22
[stm32f0xx] Add timer capture driver.

konkers:
Right now only capture is supported with the timers.  As more modes
are supported, the lower level bits of the driver should be
refactored.


konkers:
Updated with a bug fix for tim3's IRQ and support to read the counter.

travisg:
LGTM

PR #214, opened 2017-12-22
kernel/port.c: make make_buf() safe

redpig:
make_buf() is a static function with two callers that pass in one of
two constants. The parameters allowed for any uint count to be
supplied and the arithmetic during memory allocation is unchecked.

This change updates make_buf() prototype to reflect its usage, which is
safe, while eliminating any future uses from accidentally becoming
unsafe. It does so by changing the parameter as a boolean that decides
between a big or a small port buffer.

### DIFF
--- a/kernel/port.c
+++ b/kernel/port.c
@@ -91,8 +91,9 @@ typedef struct {
 static struct list_node write_port_list;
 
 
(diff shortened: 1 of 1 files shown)


travisg:
LGTM

PR #215, opened 2017-12-22
port_create race test and fix

redpig:
port_create() has a known race condition in name reservation because of the separation of name uniqueness validation and name addition.  This pull request adds a test that verifies the race is reachable, fixes the race, and updates the test with slight changes to support it.

### DIFF
--- a/app/tests/port_tests.c
+++ b/app/tests/port_tests.c
@@ -335,7 +335,12 @@ static int race_thread(void *arg)
       }
 
       port_t race_port;
-      st = port_create("racer_port", PORT_MODE_UNICAST, &race_port);
+      while(true) {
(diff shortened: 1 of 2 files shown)


app/tests/port_tests.c
```diff
@@ -291,6 +291,244 @@ static int ping_pong_thread(void *arg)
     return __LINE__;
 }
 
+const char *kStatusPortNames[] = {
```
travisg: make sure all of these are static so the compiler/linker will strip them if not building with the console.

redpig: Reposting with that change now -- thanks!

redpig: (Some of the functions in there can be made static too and I can make/test that for this pull or just do a second. Whichever works!)

PR #219, opened 2018-01-16
Pr/cdc console

konkers:
Adds the ability to get the lk console over usb cdc serial.


konkers:
Apologies for the noise.  Should be ready for review.

PR #221, opened 2018-01-28
[stm32f0xx] Correct register access for EXTI controller

teisenbe:
A bunch of the register access here didn't match STM's docs.  The old
code had a few bugs:

1) The EXTI4-15 handler would clear pending bits for interrupts 16-31.
2) GPIO interrupt configuration was very misindexed.

### DIFF
--- a/platform/stm32f0xx/exti.c
+++ b/platform/stm32f0xx/exti.c
@@ -128,7 +128,7 @@ void stm32_EXTI2_3_IRQ(void) {
 void stm32_EXTI4_15_IRQ(void) {
     arm_cm_irq_entry();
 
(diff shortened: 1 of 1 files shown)


platform/stm32f0xx/exti.c
```diff
@@ -154,11 +154,11 @@ void stm32_setup_ext_interrupt(int interrupt, stm32_ext_interrupt_port_t port,
                                bool rising_edge, bool falling_edge) {
     assert(0 <= interrupt && interrupt <= 15);
```
konkers: stm32_rcc_set_enable(STM32_RCC_CLK_SYSCFGCOMP, true) is also needed here in case the peripheral is not already enabled.

PR #224, opened 2018-03-13
Initial support for Khadas VIM2 boards

mpayam:
Basic support for VIM2
Tested on both VIM2 and qemu. 
Added support for GIC SPI based on Fuchsia gic-v2 code

### DIFF
--- a/dev/interrupt/arm_gic/arm_gic.c
+++ b/dev/interrupt/arm_gic/arm_gic.c
@@ -234,6 +234,36 @@ static int arm_gic_max_cpu(void)
     return (GICREG(0, GICD_TYPER) >> 5) & 0x7;
 }
 
+static status_t gic_configure_interrupt(unsigned int vector,
+                                        enum interrupt_trigger_mode tm,
+                                        enum interrupt_polarity pol)
+{
+    //Only configurable for SPI interrupts
+    if ((vector >= MAX_INT) || (vector < GIC_BASE_SPI)) {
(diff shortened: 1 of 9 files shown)


dev/interrupt/arm_gic/arm_gic.c
```diff
@@ -219,7 +219,8 @@ static void arm_gic_resume_cpu(uint level)
         dprintf(SPEW, "%s: distibutor is off, calling arm_gic_init instead\n", __func__);
         arm_gic_init();
         resume_gicd = true;
-    } else {
```
travisg: nit: leave this the way it was

dev/interrupt/arm_gic/arm_gic.c
```diff
@@ -250,6 +281,12 @@ void arm_gic_init(void)
         }
     }
 
+    // Initialize all the SPIs to edge triggered
+    for (i = 32; i < MAX_INT; i++) {
+        gic_configure_interrupt(i, IRQ_TRIGGER_MODE_EDGE, IRQ_POLARITY_ACTIVE_HIGH);
```
travisg: hmm, this may be problematic for other boards, but we can leave it in for now.

dev/interrupt/arm_gic/arm_gic.c
mpayam: sure. this happened once I ran the codestyle.space script. I will revert.

PR #226, opened 2018-03-22
[arm64] Add support for chain_load for arm64 platforms

mpayam:
Tested on VIM2 by loading LK from LK.

This is phase 1. Will need to add support to go back to initial boot exception level (as opposed to booting the next stage in EL1).

### DIFF
--- a/arch/arm64/arch.c
+++ b/arch/arm64/arch.c
@@ -31,7 +31,9 @@
 #include <lk/init.h>
 #include <lk/main.h>
 #include <platform.h>
+#include <target.h>
 #include <trace.h>
+#include <kernel/vm.h>
 
 #define LOCAL_TRACE 0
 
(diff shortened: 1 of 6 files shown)


travisg:
Closing this in favour of #526, which records what's actually left to do.

To be clear about why: this is *not* superseded. `arch_chain_load()` is still `PANIC_UNIMPLEMENTED` on arm64 today, so the gap you were filling is real and open. But the branch is 1550 commits behind and no longer cherry-picks cleanly, and re-reading it against your own note about phase 2, the remaining work is the hard part rather than a rebase.

Specifically, `arm_chain_load` hardcodes `sctlr_el1` with the comment "LK runs in EL1", clears `SCTLR.M` with an `isb` but no `dsb` or TLB invalidate, and leaves `SCTLR.C`/`SCTLR.I` alone. Getting back to the boot exception level — the phase 2 you flagged — isn't symmetric with `arm64_elX_to_el1` dropping down, and needs either a stashed return path or firmware help. The scratch-aspace trick you used to keep the loader identity mapped across the MMU-off transition is the right idea and I've written it into the issue so it isn't lost.

Thanks for the original work, and sorry it sat unreviewed for this long.

PR #229, opened 2018-04-15
Add stm32f070 support

teisenbe:
This cleans up handling of a lot of optional components
across f0xx models.

### DIFF
--- a/platform/stm32f0xx/vectab.c
+++ b/platform/stm32f0xx/vectab.c
@@ -78,28 +78,46 @@ DEFAULT_HANDLER(USB_IRQ)
 /* appended to the end of the main vector table */
 const void *const __SECTION(".text.boot.vectab2") vectab2[] = {
     VECTAB_ENTRY(WWDG_IRQ),                 // Window WatchDog Interrupt
+#ifdef PVD_IRQ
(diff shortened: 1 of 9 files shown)


platform/stm32f0xx/vectab.c
```diff
@@ -78,35 +78,55 @@ DEFAULT_HANDLER(USB_IRQ)
 /* appended to the end of the main vector table */
 const void *const __SECTION(".text.boot.vectab2") vectab2[] = {
     VECTAB_ENTRY(WWDG_IRQ),                 // Window WatchDog Interrupt
+#ifdef PVD_IRQ
     VECTAB_ENTRY(PVD_IRQ),                  // PVD through EXTI Line detect Interrupt
```
travisg: on each of these what happens if it's missing? does it need an empty entry in the else clause?

teisenbe: It looks like it'll try to execute at 0x0.  But these vectors don't exist on the hardware where the entries aren't defined, so they shouldn't fire.  The table as-is already has missing entries, since I think konkers wrote it against the supported interrupts on the stm32f072xb

travisg: Oh duh, I remember that whole vectab thing, it zero fills it and then uses the GCC C trickery to sparsely fill it in. Sounds good.

PR #231, opened 2018-05-09
[stm32f0xx] Fix UART3/4 interrupts

teisenbe:
The IRQ handler was being misdefined, so the correct ISR was not being
invoked.

### DIFF
--- a/platform/stm32f0xx/uart.c
+++ b/platform/stm32f0xx/uart.c
@@ -222,14 +222,14 @@ void stm32_USART2_IRQ(void)
 #endif
 
 #ifdef ENABLE_UART3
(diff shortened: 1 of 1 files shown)


teisenbe:
Note that only one of UART3 or UART4 can be enabled at a time currently (there is a #if/#error check against it earlier in the file)

konkers:
Lgtm

tfmiltz:
tyvm

travisg:
for future reference, try to use the rebase button on these. The default merge button generates an extraneous merge commit which generally just messes up the history.

teisenbe:
Oh whoops didn't even notice the different buttons. Will take a look in the
future, sorry about that

PR #233, opened 2018-06-08
fix the bug which cause memory leak in novm

zll5267:
I think this fix is needed


travisg:
Rebased and applied, thanks!

PR #234, opened 2018-09-12
[arch][arm64] Fix barriers in startup code

antonio-nino-diaz-arm:
After a TLBI instruction the right thing to do is to execute DSB followed by ISB. DSB ensures that the TLBI is seen by all observers of the system and ISB ensures that the DSB has finished before continuing.


travisg:
Ah thanks for the fix.

antonio-nino-diaz-arm:
No problem. Now that I think of it, probably arm32 also needs a similar fix...

PR #235, opened 2018-09-15
[arch][arm] Fix execution of TLBIALL instructions

antonio-nino-diaz-arm:
After a TLBI instruction the right thing to do is to execute DSB followed by ISB. DSB ensures that the TLBI is seen by all observers of the system and ISB ensures that the DSB has finished before continuing.

Also, the value in `<Rt>` is ignored for a TLBIALL. It isn't needed to load 0 to r12 and then use r12 in the instruction. In order not to cause confusion it is better to not load anything to r12.

### DIFF
--- a/arch/arm/arm/start.S
+++ b/arch/arm/arm/start.S
@@ -282,9 +282,9 @@ arm_reset:
        r8 == final translation table physical (if using trampoline)
     */
 .Lmmu_setup:
-    /* Invalidate TLB */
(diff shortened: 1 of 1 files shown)


antonio-nino-diaz-arm:
This is the counterpart of https://github.com/littlekernel/lk/pull/234 for arm32.

travisg:
looks good, thanks!

PR #255, opened 2019-11-18
[platform][tms570-launchpad] Skeleton of port to TI LaunchXL2-TMS57012

vsrinivas:
* Adds target tms570-launchpad, for TI TMS570 Launchpad Dev Kit
(https://www.ti.com/store/ti/en/p/product/?p=LAUNCHXL2-TMS57012)

* Adds CPU definitions for Cortex R4F (BE) CPU, implementing
  ARMv7-R ISA. Does not yet add definitions for ARMv7 arch entry
  functions.

* Board does not yet build - platform.c/uart.c are empty, no GIC
  entry points provided.

### DIFF
--- a/arch/arm/toolchain.mk
+++ b/arch/arm/toolchain.mk
@@ -105,6 +105,13 @@ endif
 ifeq ($(ARM_CPU),arm1176jzf-s)
 ARCH_arm_COMPILEFLAGS += -mcpu=$(ARM_CPU)
 endif
+ifeq ($(ARM_CPU),cortex-r4f)
+ARCH_arm_COMPILEFLAGS += -march=armv7-r
+ARCH_arm_COMPILEFLAGS += -mcpu=$(ARM_CPU)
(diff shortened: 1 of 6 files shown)


arch/arm/toolchain.mk
```diff
@@ -105,6 +105,13 @@ endif
 ifeq ($(ARM_CPU),arm1176jzf-s)
 ARCH_arm_COMPILEFLAGS += -mcpu=$(ARM_CPU)
 endif
+ifeq ($(ARM_CPU),cortex-r4f)
+ARCH_arm_COMPILEFLAGS += -march=armv7-r
+ARCH_arm_COMPILEFLAGS += -mcpu=$(ARM_CPU)
+ARCH_arm_COMPILEFLAGS += -mbig-endian
```
travisg: may eventually want to break the big endian switch into a separate thing, but for now this will work since there are no other R class targets.

PR #258, opened 2020-01-11
[riscv] Add support for SiFive Unleashed board, mp kernel start, and running in supervisor mode

eberman-quic:
- Adds targets for the SiFive HiFive Unleashed board and the qemu sifive-u virtual machine
- Exposes RISC-V architecture support for lk/mp kernel
  - Add signal in asm start to force secondary harts to wait for bss to be cleared.
  - Use mhartid in arch_curr_cpu_num, PLIC, and CLINT
  - Use tp register as thread pointer instead of global variable.
  - Support sending IPIs between harts using CLINT
  - Add spinlock implementation
- Adds support for running lk kernel in either supervisor or machine modes (determined by RISCV_MODE Make variable.
  - Macro-ify CSR access to use correct CSR # or use SBI call as req'd
  - Add support to make SBI calls
  - Split CLINT and lk timer abstraction so that RISC-V timer can use SBI as required.
  - Add support for booting another hart as primary since hart0 on U540 does not support S-mode. A map is used to get LK cpu number from hartid.

Got to LK test shell in follow scenarios:
- [x] SiFive HiFive Unleashed board
  - [x] Single core+Machine mode
  - [x] Single core+Supervisor mode
  - [x] Multicore+Machine mode
  - [x] Multicore+Supervisor mode
- [x] QEMU sifive-u machine
  - [x] Single core+Machine mode
  - [x] Single core+Supervisor mode
  - [x] Multicore+Machine mode
  - [x] Multicore+Supervisor mode


travisg:
Super cool! I started down the route of getting LK running in supervisor mode and using SBI but ran out of steam halfway through it. This is a cleaner implementation anyway. I'll take a look tonight and probably merge as is.

travisg:
Looks good. There's a little compile break on the sifive-e board, but it's an easy fix that I'll throw on top.

travisg:
Do you have a good way to reproduce this on Qemu? I'd like tog et that added to the scripts/do-qemuriscv script. It'll probably require booting with SBI but I've fiddled with that in the past.

eberman-quic:
> Do you have a good way to reproduce this on Qemu? I'd like tog et that added to the scripts/do-qemuriscv script. It'll probably require booting with SBI but I've fiddled with that in the past.

Sure, I can take a look.

travisg:
Have you otherwise been testing on real hardware? I am looking to get a
hodl of one of the unleashed boards, but it'll be a few weeks out at the
earliest.

eberman-quic:
Yes, I have a SiFive HiFive Unleashed board I test on as well as the few 64-bit RISC-V QEMU virtual machines.

PR #259, opened 2020-01-14
initial port of little-kernel to vc4

cleverca22:
some parts are copy/pasted from https://github.com/cleverca22/rpi-open-firmware
which itself is a fork of https://github.com/christinaa/rpi-open-firmware

what works:
network booting lk.elf on an rpi4 (just rename to start4.elf and host over tftp)
the full console with line editing and history
serial over pl011 (but no serial init, relies on previous bootloader stage)
reboot command
irq handling for timer compare and uart
one-shot timer
reporting cpu exceptions such as divide by zero


travisg:
Closing this as unmergeable at this distance, with regret — a working vc4 port that network boots on an rpi4 with console and irq handling is genuinely interesting, and "renamed to `start4.elf` and served over tftp" is a fun way to boot LK.

The problem is scale against age: +13.5k lines across 100 files, much of it derived from rpi-open-firmware, sitting five years behind a tree that has since changed the module system, the VM layer and the arch interface underneath it. Nobody here can review or maintain a new architecture backend that can't be built or tested in CI, and there's no upstream path that starts from rebasing this.

If you're still running it, the parts most likely to survive a port forward are the vc4 arch backend and the platform init, separated from the copied firmware code. Happy to look at that as a fresh PR.

travisg:
Heh sorry, the AI agent was a bit harsh, but it would be interesting to rebase the vc4 stuff on mainline.

PR #260, opened 2020-01-21
[scripts][riscv] Add support for supervisor mode in RISC-V QEMU script

eberman-quic:
Add support for supervisor mode in RISC-V QEMU script per request in #258.

Also needed to add support qemu-virt-riscv running in supervisor mode by conditionally changing MEMBASE to 0x80200000 so it does not overlap with OpenSBI.

Line ref for QEMU RISC-V "any" core not having supervisor mode: https://github.com/qemu/qemu/blob/master/target/riscv/cpu.c#L116

### DIFF
--- a/platform/qemu-virt-riscv/rules.mk
+++ b/platform/qemu-virt-riscv/rules.mk
@@ -13,7 +13,11 @@ MODULE_SRCS += $(LOCAL_DIR)/plic.c
 MODULE_SRCS += $(LOCAL_DIR)/uart.c
 
 #ROMBASE ?= 0x20400000 # if running from rom, start here
-MEMBASE ?= 0x80000000
+ifeq ($(RISCV_MODE),supervisor)
+MEMBASE ?= 0x080200000
+else
+MEMBASE ?= 0x080000000
+endif
 MEMSIZE ?= 0x00100000 # default to 1MB
 
 # sifive_e or _u?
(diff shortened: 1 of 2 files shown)


eberman-quic:
I only added support for supervisor mode in that script, not for the sifive-u machine. I wasn't sure how you wanted to do the options for that as sifive-u, sifive-e, and virt all run on 32- and 64-bit. Probably could do `:uev:` for the 3 platforms and enforce mutual exclusivity.

travisg:
Yah, actually I ended up doing basically the same thing in another branch that i'm about to merge into mainline. Mainly what I'm interested in doing is finding a way to test the Sifive-U target in qemu, which I think it has support for? machine=virt is very easy to test and my main strategy.

Going to work on adding MMU support momentarily.

travisg:
Though I wont take this one I'll make sure its usage maps into my slightly different variant of it.

Thanks, really appreciate the work. It's gotten me interested in the platform again.

eberman-quic:
> Mainly what I'm interested in doing is finding a way to test the Sifive-U target in qemu, which I think it has support for? machine=virt is very easy to test and my main strategy.

This is the command I usually run:
```
qemu-system-riscv64 -M sifive_u -serial stdio -display none -kernel build-qemu-sifive-u-test/lk.elf -bios none
```

PR #261, opened 2020-05-14
Migrating from StellarisWare to TivaWare

CunningLearner:
This patch supports new platforms available under same TivaWare umbrella,
enhances portability and correct architectural deficiencies and inconsistencies
in StellarisWare.

* Add latest TivaWare_C_Series driverlib
* Replace existing LM4F120H5QR with its relative part TM4C1233H6PM

More detailed features for further migration can be referenced from:
http://www.ti.com/lit/an/spma050a/spma050a.pdf

Tested on Tiva C series TM4C123GH6PM.

Signed-off-by: Himanshu Sahdev aka CunningLearner <sahdev.himan@gmail.com>


travisg:
Oh very nice! I'll have to try it out on my old tiva boards. I do remember TI renaming everything while I still had my stellaris board.

travisg:
This looks nice, but there is a built failure or two in there, can you take a look? I have my old Tiva C launchpad ready to test this on.

CunningLearner:
> This looks nice, but there is a built failure or two in there, can you take a look? I have my old Tiva C launchpad ready to test this on.

Yeah, the problem is with the toolchain `arm-eabi-7.3.0-Linux-x86_64` I guess, I am using `arm-none-eabi-7-2018-q3-update` and it works fine! Also check using their respective higher versions available.

**driverlib/hibernate.c** needs **time.h** which is not included in the `arm-eabi` but is available with `arm-none-eabi`. travis-build only uses `arm-eabi` and throws error

1. Removing hibernate.c from MODULE_SRCS works well. So we can push initially without including it in rules.mk `You can also test it initially`
2. Both StellarisWare and TivaWare looks for `arm-stellaris-eabi` if it exists, fallback to `arm-none-eabi` otherwise. Can we use arm-none-eabi for travis-build as well?
3. ?

travisg:
Closing this one out. For the record, your diagnosis was almost certainly right — `driverlib/hibernate.c` wanting `time.h`, which `arm-eabi` doesn't ship but `arm-none-eabi` does, is a real toolchain difference and not a defect in your patch. It stalled on our side after that, not yours.

What makes it unmergeable now is the shape rather than the content: it vendors the full TivaWare C Series driverlib, which is over a million lines, into the tree. LK doesn't take vendor SDK imports at that size any more — the `external/` policy has tightened considerably since 2020 — and the branch is now years behind the platform and module layers it plugs into.

The underlying point stands: TI renamed everything and `platform/stellaris` is against the old StellarisWare names. If someone wants Tiva support, the way in is a `platform/tiva` that pulls in only the driverlib pieces it actually uses. Thanks for doing the work, and sorry it didn't get resolved when you were still on it.

PR #263, opened 2020-05-17
Fix incompatibility with Make 4.3.

lgorence:
The behavior of += has changed such that it will not prepend a space, if the original variable was empty. This breaks the generation of config.h. The fix I've found is to create the variable with two empty variables and a space between, and this seems to work for both Make 4.2 and 4.3.

Source: https://lwn.net/Articles/810071/
```
* WARNING: Backward-incompatibility!
  Previously appending using '+=' to an empty variable would result in a value
  starting with a space.  Now the initial space is only added if the variable
  already contains some value.  Similarly, appending an empty string does not
  add a trailing space.
```


travisg:
Oh thanks, the new make 4.3 features look good. Will roll this in a sec.

PR #264, opened 2020-07-17
[stm32f0xx] Fix prescaler value.

claridge:
As specified in ST reference manual RM0091, the prescaling factor
applied by timers is 1 plus the value in the corresponding prescaler
register. (See section 17.4.11 for TIM1, and analogous sections for
other timers.)

This change causes stm32_timer_capture_setup to take the prescaling
factor as input rather than the register value.

### DIFF
--- a/platform/stm32f0xx/timer_capture.c
+++ b/platform/stm32f0xx/timer_capture.c
@@ -316,6 +316,8 @@ void stm32_TIM17_IRQ(void) {
 }
 
 status_t stm32_timer_capture_setup(stm32_timer_capture_t *tc, int timer, uint16_t prescaler) {
+    assert(prescaler > 0);
+
     tc->config = stm32_timer_get_config(timer);
     if (tc->config == NULL) {
         return ERR_NOT_FOUND;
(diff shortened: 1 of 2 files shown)


platform/stm32f0xx/timer_capture.c
```diff
@@ -329,7 +331,9 @@ status_t stm32_timer_capture_setup(stm32_timer_capture_t *tc, int timer, uint16_
     spin_lock_init(&tc->overflow_lock);
 
     tc->config->regs->CR1 = 0;
-    tc->config->regs->PSC = prescaler;
+
+    // The prescaler value applied by hardware is PSC + 1.
+    tc->config->regs->PSC = prescaler > 0 ? prescaler - 1 : 0;
```
konkers, line 336: Curious why we have both the assert for prescaler > 0 at the top of the function as well as the check here.

claridge: It's a habit I got into at Makani -- assert on your expectations, but then try to do something reasonable if they fail in case the code is compiled without asserts.

travisg: Yah I generally do the same thing. LGTM.

claridge: Thanks, Travis!

PR #266, opened 2020-07-21
[kernel][thread] Wake up only needed cores

vishals4gh:
This change ensures that during thread scheduling
only needed cores are woken up as per the thread's
CPU affinity rather than all cores. This would avoid
un-necessary wakup for other cores from lower power
states without having to execute anything.


travisg:
Looks good. Thanks! Sorry it took so long to get to it, been busy with other stuff.

PR #267, opened 2020-07-22
[ARCH][ARM64] Dump more information during aborts

vishals4gh:
1) Decode FSC and dump more human readable status
2) Add support of stack unwinding as referred from
arm64 procedure call standard and frame pointer usage.
3) Compiler options for not omitting frame pointer
are enabled to ensure usage of frame pointers even
with higher optimization levels enabled.


travisg:
Looks good! I'll take a look at these changes tomorrow but I definitely appreciate the good work here.

arch/arm64/rules.mk
```diff
@@ -91,6 +91,7 @@ TOOLCHAIN_PREFIX := $(ARCH_$(ARCH)_TOOLCHAIN_PREFIX)
 $(info TOOLCHAIN_PREFIX = $(TOOLCHAIN_PREFIX))
 
 ARCH_COMPILEFLAGS += $(ARCH_$(ARCH)_COMPILEFLAGS)
+ARCH_COMPILEFLAGS += -fno-omit-frame-pointer -fno-optimize-sibling-calls
```
travisg, line 94: Not so sure about these, was this accidentally included? Omit frame pointer I can see, but the optimize sibling calls may be kind of a performance hit?

arch/arm64/exceptions_c.c
```diff
@@ -18,6 +18,123 @@ struct fault_handler_table_entry {
     uint64_t fault_handler;
 };
 
+struct fault_status_map {
+    uint32_t fsc;
+    const char *fault_msg;
+};
+
+/* Instruction and Data abort share the fault status encoding */
+static struct fault_status_map fsc_map[] = {
```
travisg, line 27: should be able to mark this static const so it can live in the rodata section. compiler might be smart enough to figure that out based on usage, but probably better to be explicit.

travisg (commented):
Looks pretty good, just a few tweaks.

arch/arm64/rules.mk
vishals4gh: I had referred to Linux implementation for these flags. True, there is a performance hit though this could allow pinpointing the recursion depth properly with this flag. Will remove this one since its not too critical to have this option passed to the compiler.

arch/arm64/exceptions_c.c
vishals4gh: Ack.

PR #268, opened 2020-09-09
[platform][tms570-launchpad] Do not try to save d16-d31 on ARMv7-R

vsrinivas:
All shipping ARMv7-R processors include VFPv3-D16, a subset of VFPv3
with only 16 double-precision registers. The first -R profile CPU
with a complete VFP or NEON implementation is the Cortex-R52,
implementing ARMv8-R.

LK's context switch code had a dynamic check and would only save
d16-d31 if present, but gcc/gas will not assemble code that includes
references to d16-d31 when mcpu=cortex-r4f or other v7-R CPUs.

Ideally we'd key the #if off of __TARGET_FPU_VFPV3_D16, but that
only appears to be defined by the ARM compilers, not gcc. Use
V7-R as the key instead.

### DIFF
--- a/arch/arm/arm/fpu.c
+++ b/arch/arm/arm/fpu.c
@@ -82,9 +82,11 @@ void arm_fpu_thread_swap(struct thread *oldthread, struct thread *newthread) {
 
             __asm__ volatile("vmrs  %0, fpscr" : "=r" (oldthread->arch.fpscr));
             __asm__ volatile("vstm   %0, { d0-d15 }" :: "r" (&oldthread->arch.fpregs[0]));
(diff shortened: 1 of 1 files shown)


vsrinivas:
Part 1 of fixing the TMS570 build; the second part is to remove the GIC rules and add VIM support. Working on that now.

travisg (approved):
LGTM. Probably can do something slightly more fancy if/when a armv8-r ever gets in there, but probably would be 64bit only so no problem.

travisg:
Feel free to add this target to the travis-ci build as well so it gets tested. It's the only armv7-r in the matrix.

Issue #272, opened 2020-10-11
clean up PL011 duplicating


cleverca22:

travisg:
Yah good idea. There might even be a third one.

A similar uart that's duplicated a few times is 8250 serial stuff.

PR #274, opened 2020-10-11
[dev][pl011] fix issue #272

cleverca22:
while merging the 2 drivers, i found a bug in the qemu variant of `uart_getc` that didnt reveal itself under qemu, and fixed it

i also added some code for the other 4 PL011's in the pi4, but no target for that yet


travisg:
Nice, looks like a good start. I'll do a review in a little bit but on pass, can you update the comment itself to include what it's doing? I'm not a fan at all of commit changes that just say it fixes a commit # without any context.

Something like [dev][uart][pl011] Move to common driver

And then a line or two describing what it did, including the bug fix and a mention that it fixes a bug #.

cleverca22:
i do have some further improvements to the uart driver on the vc4 fork, but trying to keep this commit clear of those, so its easier to see just the merging, and the improvements can come later

dev/pl011/include/dev/pl011.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
```
travisg, line 1: Please add the standard license header here.

dev/pl011/uart.c
```diff
@@ -37,24 +37,22 @@
...
-        case 0:
-            return UART_BASE;
-    }
+    return uart_base[n];
 }
 
 static enum handler_return uart_irq(void *arg) {
     bool resched = false;
-    uint port = (uintptr_t)arg;
+    uint port = (uint)arg;
```
travisg, line 48: This is an issue with 64bit targets, since its down casting to a smaller size. Perhaps cast it through uintptr_t first, then down to uint.

dev/pl011/uart.c
```diff
@@ -83,37 +81,38 @@ static enum handler_return uart_irq(void *arg) {
     return resched ? INT_RESCHEDULE : INT_NO_RESCHEDULE;
 }
 
-void uart_init(void) {
-    for (size_t i = 0; i < NUM_UART; i++) {
-        uintptr_t base = uart_to_ptr(i);
+void pl011_uart_init(int irq, int nr, uintptr_t base) {
```
travisg, line 84: To generally be consistent with existing apis, put the number first.

dev/pl011/uart.c
```diff
@@ -83,37 +81,38 @@ static enum handler_return uart_irq(void *arg) {
...
+    assert(nr < NUM_UART);
+    uart_base[nr] = base;
+    // create circular buffer to hold received data
+    cbuf_initialize(&uart_rx_buf[nr], RXBUF_SIZE);
 
-        // create circular buffer to hold received data
-        cbuf_initialize(&uart_rx_buf[i], RXBUF_SIZE);
+    // assumes interrupts are contiguous
+    register_int_handler(irq, &uart_irq, (void *)nr);
```
travisg, line 91: the (void *) is a warning on arm64. Perhaps as before, cast it through uintptr_t first.

platform/qemu-virt-arm/rules.mk
```diff
@@ -33,6 +32,7 @@ MODULE_DEPS += \
     dev/virtio/block \
     dev/virtio/gpu \
     dev/virtio/net \
+    dev/pl011 \
```
travisg, line 35: To be consistent, sort the list of the devs.

platform/bcm28xx/rules.mk
```diff
@@ -42,8 +42,8 @@ SMP_CPU_ID_BITS := 8
 GLOBAL_DEFINES += \
 	BCM2836=1
 
-MODULE_SRCS += \
-	$(LOCAL_DIR)/uart.c
+MODULES += \
```
travisg, line 45: Use MODULE_DEPS instead of MODULES. MODULES sometimes works but the _DEPS is the correct solution.

dev/pl011/rules.mk
```diff
@@ -0,0 +1,7 @@
+LOCAL_DIR := $(GET_LOCAL_DIR)
```
travisg, line 1: Please name this dev/uart/pl011 instead of at the top level of dev.

travisg (changes requested):
A few comments, but nothing majorly structurally wrong.

PR #276, opened 2020-10-14
[arm][gic] Ensure context switch doesn't happen from irq

vishals4gh:
Context switches should not happen from within the interrupt
context before interrupt is cleared by write to GIC EOIR register,
without it GIC will simply keep that interrupt active even if
the hardware source clears the interrupt to the gic, causing
subsequent irqs from the source to not get delivered to the CPU.

This change adds an assertion that context switch doesn't happen
from irq context before interrupt is EOIed. TCB field is added
to convey if the current thread has interrupt context active, if
so thread_resched should ideally not get called.

### DIFF
--- a/.github/workflows/notify-irc.yml
+++ b/.github/workflows/notify-irc.yml
@@ -11,6 +11,7 @@ jobs:
         with:
           channel: "#lk"
           nickname: lk-github
+          notice: true
           message: |
             ${{ github.actor }} pushed ${{ github.event.ref }} ${{ github.event.compare }}
             ${{ join(github.event.commits.*.message) }}
@@ -20,6 +21,7 @@ jobs:
         with:
           channel: "#lk"
           nickname: lk-github
+          notice: true
           message: |
             ${{ github.actor }} opened PR ${{ github.event.html_url }}
       - name: irc tag created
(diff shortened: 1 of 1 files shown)


kernel/include/kernel/thread.h
```diff
@@ -106,6 +106,10 @@ typedef struct thread {
     /* thread local storage */
     uintptr_t tls[MAX_TLS_ENTRY];
 
+    /* Flag to indicate that interrupt
+     * context is active for the thread */
+    uint32_t intr_ctxt_active;
+
     char name[32];
```
vishals4gh, line 113: Probably this could have been better handled by per CPU interrupt state.

travisg: Also problematic is it is only implemented on ARM. For a new feature like this it really should be implemented on all of the architectures, at least if possible.

IIRC, there's an arch_in_interrupt_handler() routine that may be actually what you're looking for. Or at least having it be an arch_ routine that can be stubbed out by default if the arch doesn't want to implement it. I'm leery about adding state at the thread level.

travisg: Hmm, actually looks like the arch_in_int_handler() is only basically implemented on cortex-m, since the 'big arm' one implements it as a global bool which is clearly not SMP safe. Looks like might make sense to fix that first.

vishals4gh: Thanks Travis for the feedback, makes sense. I will try to upload something for all the archs, assuming no arch wants to allow context switching to happen between IRQ handling.

PR #280, opened 2020-10-25
[arm] Ensure context switch doesn't happen from irq

vishals4gh:
Context switches should not happen from within the interrupt
context before interrupt is cleared by write to GIC EOIR register,
without it GIC will simply keep that interrupt active even if
the hardware source clears the interrupt to the gic, causing
subsequent irqs from the source to not get delivered to the CPU.

This change adds an assertion that context switch doesn't happen
from irq context before interrupt is EOIed. TCB field is added
to convey if the current thread has interrupt context active, if
so thread_resched should ideally not get called.

Signed-off-by: vannapurve vannapurve@google.com

### DIFF
--- a/arch/arch.c
+++ b/arch/arch.c
@@ -1,1 +1,13 @@
-// empty file to help build empty arch module
+/*
+ * Copyright (c) 2020 Travis Geiselbrecht
+ *
+ * Use of this source code is governed by a MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT
+ */
+
+#include <arch/ops.h>
+
+__WEAK bool arch_in_int_handler(void) {
+    return false;
+}
--- a/arch/arm/arm/arch.c
+++ b/arch/arm/arm/arch.c
@@ -47,6 +47,32 @@ static void arm_basic_setup(void);
 static void spinlock_test(void);
 static void spinlock_test_secondary(void);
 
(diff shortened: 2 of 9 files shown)


vishals4gh:
Accidentally closed https://github.com/littlekernel/lk/pull/276 when trying to update my branch with mainline.

Please assume the discussion to be in continuation of the above pull request. I have uploaded this patch to update support for just arm/arm64 archs for now. If this scheme looks ok, I can extend the support to other archs.

Alternative scheme I was thinking about was some generic layer like kerne/thread exposing APIs to the bookkeeping of CPU IRQ context which can be called from arch specific exception handlers, this can reduce code duplication in all the archs.

travisg:
Closing this in favour of #528, which records the invariant and the open design question.

The bug you're describing is real — resched before the `GICC_EOIR` write leaves the interrupt active at the GIC, and the source goes deaf afterwards. That's worth an assertion at minimum.

On the alternative you raised in the comments: putting the bookkeeping behind a small `kernel/thread.c` API that arch exception handlers call, rather than duplicating per-arch state, is very likely the right shape. The invariant isn't arm-specific even though the GIC makes the failure obvious. Worth knowing before anyone starts: #368 reports that `arch_in_int_handler()` is broken on multicore anyway — `__arm_in_handler` is a single global rather than per-cpu — so the existing hook may be part of what needs fixing rather than something to build on.

Also noting #276 was closed accidentally and this was its continuation; neither got a proper reply at the time, which is on us. Thanks for the analysis.

PR #293, opened 2021-01-25
[libc] Fix UINT16_MAX definition.

tworaz:
The name of the define provided by limits.h is USHRT_MAX not USHORT_MAX.


travisg:
Oh yeah you're right. Also looks like INT16_MAX needs to be fixed to (SHORT_MAX). Can you revise?

tworaz:
INT16_MIN suffered from the same problem. The latest version should fix all of those.

PR #298, opened 2021-05-08
[arch][x86] Unset x86 inner page table entry G bit

eastmoutain:
The issue was initially found on AMD machine when run lk with qemu kvm, it
can't boot if kvm hardware is enable in qemu.

According to Intel system programming guild Chapter 4 "Paging",
if the page table entry is non-leaf entry, then the G bit will be
ignored.

However, According to AMD programmer mannul Volume 2, Chapter 5.3
"Long-Mode Page Translation", the non-leaf page table entry G bit
must be zero.

The patch clears inner page table entry G bit to zero so that it
works on both Intel and AMD CPU

### DIFF
--- a/arch/x86/include/arch/x86/mmu.h
+++ b/arch/x86/include/arch/x86/mmu.h
@@ -26,7 +26,7 @@
 
 #ifndef X86_LEGACY
 /* default flags for inner page directory entries */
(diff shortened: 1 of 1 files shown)


travisg:
Oh thanks! I even knew about this one and forgot to apply it. Shows you how often I test x86, given that i'm largely AMD based.

travisg:
Yeah, I discovered it in the Fuchsia fork of this and fixed it there. Thought I had rolled it back but guess I had forgotten.

PR #299, opened 2021-06-03
[arch][arm64] unmask_interrupt needs the same numbers as register_int_handler

jatron:
[arch][arm64] unmask_interrupt needs the same numbers as register_int_handler


jatron:
@travisg , is this something that can be merged?

travisg:
Sure! It's commented out but makes sense to keep it in sync.

PR #303, opened 2021-08-20
initial code to support ext4

cleverca22:
currently lacks the ability to deal with files with more then 4 extents
lacks the ability to deal with filesystems over 2^32 blocks long
only tested on files with 1 extent

based on reading https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout


travisg:
hey clever can you rebase this potentially?

PR #305, opened 2021-09-12
[arch][arm] improve arm chainload

cleverca22:
arch_mmu_map was failing hard, because the identity mapping does not fall within the `vmm_get_kernel_aspace`

this creates a new aspace covering the loader, so it can identity map

linux is also unable to use the FPU if lazy FPU context switching had turned it off prior to the chainload, `arm_fpu_set_enable()` is used to turn it back on


travisg:
LGTM, can you add a bit more in the comment section of the CL describing what you did?

Basically the FPU disabling and the aspace thing?

PR #308, opened 2021-11-09
Fix module.mk to use the correct include flag

mark64:
`--include` is not a real gcc flag. Maybe it exists for other compilers
or I'm just bad at reading docs though.

Anyway, now that this uses the expected gcc flag, `ccache` works when
added to this build system. Without this, `ccache` thinks there's
multiple input files and refuses to cache results. Now, it works and has
a great hit rate.


travisg:
Oh! That's interesting. I've known about ccache and --include for a long time, but didn't realize it was an improper use of the switch. My guess is gcc just accepts both forms but ccache only matches against one of them. Will test and roll. Thanks!

travisg:
Yah interesting! Gcc seems to understand both, but having --include and -include confuses ccache. Just using -include seems to do the trick. Thanks!

PR #320, opened 2022-03-30
[arch][arm64] disable EL1 FPU traps even when starting at EL1

pcc:
It is possible for early initialization functions such as lk_main()
to contain NEON instructions because we don't build the kernel with
-mgeneral-regs-only. As a result we can end up taking an FPU exception
before we are ready to handle it.

We didn't have this problem when starting at a higher exception level
than EL1 because we turned off FPU traps in arm64_elX_to_el1(). But we
neglected to do so when starting at EL1. Fix the problem by moving the
CPACR_EL1 manipulation out of arm64_elX_to_el1() and into arm_reset().

### DIFF
--- a/arch/arm64/start.S
+++ b/arch/arm64/start.S
@@ -35,6 +35,10 @@ arm_reset:
     /* if we came in at higher than EL1, drop down to EL1 */
     bl      arm64_elX_to_el1
 
+    /* disable EL1 FPU traps */
+    mov     tmp, #(0b11<<20)
+    msr     cpacr_el1, tmp
+
 #if WITH_KERNEL_VM
     /* enable caches so atomics and spinlocks work */
     mrs     tmp, sctlr_el1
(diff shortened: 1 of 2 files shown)


arch/arm64/start.S
```diff
@@ -35,6 +35,10 @@ arm_reset:
     /* if we came in at higher than EL1, drop down to EL1 */
     bl      arm64_elX_to_el1
 
+    /* disable EL1 FPU traps */
+    mov     x0, #(0b11<<20)
```
travisg, line 39: Looks good, can you switch this to using the 'tmp' register however? Reason is in some situations x0 may still contain an argument from the bootloader or from PSCI when booting secondary cpus. The registers in most of this file try to avoid using x0-x3 or so.

travisg (changes requested):

travisg (approved):
looks good, thanks!

PR #321, opened 2022-03-30
[arch][arm64] use SMC when booted at EL2 in QEMU

pcc:
If we were booted at EL2 (e.g. when passing -machine
virt,virtualization=on), we need to use SMC instead of HVC for PSCI
calls. Change psci_call() to do this and add a flag to do-qemuarm to
allow testing this scenario.

### DIFF
--- a/arch/arm64/start.S
+++ b/arch/arm64/start.S
@@ -27,11 +27,15 @@ paddr                   .req x24
 mapping_size            .req x25
 size                    .req x26
 attr                    .req x27
+boot_el                 .req x28
 
 .section .text.boot
 FUNCTION(_start)
 .globl arm_reset
 arm_reset:
+    /* keep track of the boot EL */
+    mrs     boot_el, currentel
+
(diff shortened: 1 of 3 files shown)


arch/arm64/start.S
```diff
@@ -32,6 +32,11 @@ attr                    .req x27
 FUNCTION(_start)
 .globl arm_reset
 arm_reset:
+    /* keep track of the boot EL */
```
travisg, line 35: This is tricky, since all cpus enter via this hook. Should probably only save it when on cpu 0, though probably all of the other cpus enter at the same EL so mostly likely it works. 

Can you add a branch here that tests for cpu 0 (mpidr with AFF0-3 masked off) and only saves it for that one? Or, save it into a register and then only conditionally write it out in the code later on that is known to be on cpu 0 (around line 73).

travisg (changes requested):

arch/arm64/start.S
pcc: Done.

PR #322, opened 2022-04-01
[build] make LK buildable with LLVM/Clang

pcc:
Add an environment setting LLVM=1 that replaces usage of GNU tools
with the LLVM equivalents.

Current status is that it builds on arm64 for QEMU with several
warnings and then boots. I didn't hook up any other architectures but
the arm64 setup should serve as a guideline.


travisg:
Oh I just saw this after mentioning that should add support for LLVM. Will give it a go to make sure it works on all the arches. Seems like it's just for arm64 at the moment?

pcc:
Yes, I'd expect this to fail on the other arches because it would need some changes to the other arch-specific makefiles. I'm currently only set up to build for arm64 (e.g. only have arm64 baremetal builtins library to hand) so I didn't touch the other arches.

pcc:
Ping... it would be nice to land this so that I don't have to keep carrying it around when testing other patches (I don't have a compatible GCC installation on this machine).

travisg:
Yah, will take a look. I need to resolve it against other arches and see
how to keep that going. Also have some work recently with fpu enablement
that this might intersect with.

travisg:
One thing that might help is to break this CL into two parts. One is just stuff needed to get clang to compile and the second one would be build system config stuff. The former is a slam dunk, the latter might need a bit of thinking.

pcc:
Sure, I was thinking that because this doesn't *regress* the other arches (you can still compile them with GCC) that wouldn't be a blocker to landing it at this point. Also some of the arches don't have LLVM backends at this point (microblaze, or1k) so this is probably never going to be "complete" in terms of arch support.

Would it help to have some kind of explicit error if you try to build an unsupported arch with `LLVM=1`?

travisg:
Closing this as superseded — master now builds with Clang, including the pieces this PR started.

Current state: `TOOLCHAIN=clang` is wired through `engine.mk`, `LD=ld.lld` works, and `LTO_MODE=thin|full` is supported on top of that. CI has a Clang matrix spanning versions 14, 18 and 20 at `DEBUG=2` and `DEBUG=0` across the arm32, arm64, riscv32, riscv64 and riscv64-supervisor qemu projects, plus a separate LTO spot-check job. `make help` documents all three knobs.

Thanks for pushing on this early — it's a standard part of the build now.

PR #323, opened 2022-04-06
[top] use incomplete array declarations for constructor symbols

pcc:
Clang will omit the initial comparison in this while loop on the
assumption that two different variables will not have the same address,
which will lead to a crash if the kernel has no constructors.

It will, however, retain the comparison for incomplete arrays
because they may have zero size (and therefore may alias with another
variable). Therefore, change the declarations of the start/end symbols
for the constructor list to incomplete arrays.

### DIFF
--- a/top/main.c
+++ b/top/main.c
@@ -29,8 +29,8 @@
 /* saved boot arguments from whoever loaded the system */
 ulong lk_boot_args[4];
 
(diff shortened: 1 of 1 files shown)


travisg:
Thanks! I really need to add proper support for clang in the build system as well, since a lot of projects downstream use it.

PR #324, opened 2022-04-13
[build] Use CC as a linker

heatd:
Make lk use the compiler as the linker and convert all of the passed flags into CC equivalents


travisg:
I have a locally rebased version. Looks pretty good, going to test it alongside another CL that enables clang and make sure they don't conflict.

PR #327, opened 2022-05-12
[arch][arm64] replace the trampoline translation table with a trampol…

pcc:
…ine VBAR

I noticed that LK failed to boot on systems that do not support 64KB
page sizes (e.g. Linux KVM guest on Apple M1) because the trampoline
translation table used a compile-time hardcoded 64KB page size.

Instead of trying to make the trampoline translation table code
look for a supported page size at runtime, I realized that it should
be possible to remove the trampoline translation table entirely by
replacing it with a VBAR that branches to the instruction following
the MMU enable. That's what this patch does.


travisg:
Yah I was thinking the same thing at some point, but was a little worried that some random cpu and/or emulator would freak out in that situation. May still be worth doing however.

travisg:
re: the 64k thing, had to fix that in zircon by just forcing the trampoline code to use 4K pages. limits the physical address space to 39 bits if just wanting to use two levels of page tables.

Trusty seems to have gone in the different direction and moved the trampoline (and all the early mmu page tables) into C, much like what we did for zircon. Also thinking of picking that up as well.

travisg:
Was initially a little worried about it but the more I fiddle with this patch the more I'm pretty happy with it, so I think I'll take it pretty much as is. Thanks!

PR #328, opened 2022-05-12
[arch][arm64] determine the correct TCR_EL1.IPS at runtime

pcc:
Change the early startup code to set TCR_EL1.IPS to
ID_AA64MMFR0_EL1.PARange if it has a defined value (the currently
defined values have the same meanings), but use 48-bit PAs if 52-bit
PAs are supported because 52-bit PAs have a different translation
table format that we don't support. Stash the computed TCR_EL1 in a
variable and use it in the context switch code.

### DIFF
--- a/arch/arm64/mmu.c
+++ b/arch/arm64/mmu.c
@@ -30,6 +30,9 @@ pte_t arm64_kernel_translation_table[MMU_KERNEL_PAGE_TABLE_ENTRIES_TOP]
 __ALIGNED(MMU_KERNEL_PAGE_TABLE_ENTRIES_TOP * 8)
 __SECTION(".bss.prebss.translation_table");
 
+/* the base TCR flags, computed from early init code in start.S */
+uint64_t arm64_mmu_tcr_flags __SECTION(".bss.prebss.tcr_flags");
+
 static inline bool is_valid_vaddr(arch_aspace_t *aspace, vaddr_t vaddr) {
     return (vaddr >= aspace->base && vaddr <= aspace->base + aspace->size - 1);
 }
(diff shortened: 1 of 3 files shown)


Tunealb (commented):
👍

pcc:
(Rebased to head.)

arch/arm64/start.S
```diff
@@ -266,6 +266,24 @@ arm_reset:
     b       .Lmap_range_one_table_loop
 
 .Linitial_mapping_done:
+    ldr     tmp, =MMU_TCR_FLAGS_BASE
```
travisg, line 269: suggestion: put a quick comment here as to what this block is for. Something to the effect of 'compute the base TCR configuration and save away in a global for future use.'

arch/arm64/mmu.c
```diff
@@ -30,6 +30,8 @@ pte_t arm64_kernel_translation_table[MMU_KERNEL_PAGE_TABLE_ENTRIES_TOP]
 __ALIGNED(MMU_KERNEL_PAGE_TABLE_ENTRIES_TOP * 8)
 __SECTION(".bss.prebss.translation_table");
 
+uint64_t arm64_mmu_tcr_flags __SECTION(".bss.prebss.tcr_flags");
```
travisg, line 33: suggestion: comment mentioning that this is the base TCR flags, computed from early init code in start.S

travisg (approved):
looks good with a few comments.

arch/arm64/mmu.c
pcc: Done

arch/arm64/start.S
pcc: Done

PR #335, opened 2022-07-21
F4 update

luka177:


travisg (commented):
There are two problems with these CLs.

1) The line endings are windows (CRLF) so it seems to be thinking every single line has changed. Unclear what OS you're using to check it in but if you could convert to unix or at least whatever line ending git thinks is the same as before (shoudl show up as far less diffs).
2) the +x bits are being set on every file, probably a result of editing from windows or across a SMB share, or maybe an issue in their zip file.
3) Around line 11583 in stn32f4xx.h there's a double declaration of a few items. Looks like sloppy code on ST's part but it generates a pile of warnings. We should pick one of the two.

Otherwise looks good to me.

platform/stm32f4xx/rules.mk
```diff
@@ -14,16 +14,16 @@ ARM_CPU := cortex-m4
 
 # TODO: integrate better with platform/stm32f4xx/CMSIS/stm32f4xx.h
 ifeq ($(STM32_CHIP),stm32f407)
-GLOBAL_DEFINES += STM32F40_41xxx 
+GLOBAL_CFLAGS=-DSTM32F40_41xxx=1
```
travisg, line 17: It's preferable to use GLOBAL_DEFINES += STM32F40_41xxx=1 here instead of CFLAGS. GLOBAL_CFLAGS doesn't work exactly how you think it does, and the defines tracks things better by putting in a header file instead of on the command line.

travisg (changes requested):

external/platform/stm32f4xx/STM32F4xx_StdPeriph_Driver/CMSIS/stm32f4xx.h
```diff
@@ -50,6 +50,8 @@
  extern "C" {
 #endif /* __cplusplus */
   
+#define assert_param(expr) ((void)0)
```
travisg, line 53: Can you leave a comment here as to what it's for? It's possible this a good define that we want to pull into LK itself.

travisg (changes requested):

external/platform/stm32f4xx/STM32F4xx_StdPeriph_Driver/CMSIS/stm32f4xx.h
luka177: St code depends on assert_param(expr) but it is not defined in lk.

platform/stm32f4xx/rules.mk
luka177: Interesting GLOBAL_DEFINES didn't work, do I need to include that header somehow?

external/platform/stm32f4xx/STM32F4xx_StdPeriph_Driver/CMSIS/stm32f4xx.h
luka177: https://github.com/littlekernel/lk/blob/a319a1f68d592529c03a1e95c369ffc0a2ed20cd/external/platform/stm32f0xx/STM32F0xx_HAL_Driver/stm32f0xx_ll_tim.c#L44

luka177:
Line encoding seems ok here now.

travisg:
Merged locally with a few fixes here and there. Thanks!

PR #338, opened 2022-07-22
[platform-stm32f4xx]: Implement sdram support

luka177:


travisg (approved):

travisg:
Thanks for this, and sorry it sat for so long — it was approved back in 2022 and then never landed.

I've merged it as 06b2f87b with your authorship preserved, plus three changes:

- `sdram.c` is now compiled only for `STM32_CHIP := stm32f429`. The F40x parts have an FSMC without the SDRAM controller, so as posted this broke `stm32f4-discovery-test` (`RCC_AHB3Periph_FMC`, `GPIO_AF_FMC` and `FMC_SDSR_BUSY` are all undeclared there).
- `NOVM_MAX_ARENAS=2` moved to `target/stm32f429i-disco/rules.mk`, next to `SDRAM_BASE`/`SDRAM_SIZE`/`ENABLE_SDRAM`, rather than being set for every F4 part.
- `platform/sdram.h` carried a prototype for `SDRAM_Init()`, which doesn't exist; renamed it to the real `stm_SDRAM_Init()` so `init.c` no longer needs its own extern.

Compile tested on `stm32f429i-disco-test` and `stm32f4-discovery-test`, confirming `sdram.c` is built only for the former. Not runtime tested — I don't have the board, so if you still have yours I'd appreciate a sanity check that the arena actually comes up.

Closing this in favour of the merged commit.

PR #340, opened 2022-08-04
[lib][stdio] wrap file io APIs with stdio APIs

chiehmin:
Implement stdio file io APIs with LK lib/fs APIs

Signed-off-by: Chieh-Min Wang <cmwang@google.com>

### DIFF
--- a/lib/libc/stdio.c
+++ b/lib/libc/stdio.c
@@ -6,15 +6,24 @@
  * https://opensource.org/licenses/MIT
  */
 #include <lk/debug.h>
+#include <lk/err.h>
 #include <stdio.h>
 #include <stdarg.h>
 #include <string.h>
+#include <stdlib.h>
 #include <sys/types.h>
 #include <platform/debug.h>
+
+#if defined(WITH_LIB_FS)
+#define STDIO_FIELDS .use_fs = false,
+#else
+#define STDIO_FIELDS
+#endif // WITH_LIB_FS
 
 #define DEFINE_STDIO_DESC(id)   \
     [(id)]  = {                 \
         .io = &console_io,      \
+        STDIO_FIELDS            \
     }
 
 FILE __stdio_FILEs[3] = {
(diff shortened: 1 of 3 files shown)


travisg:
Oh I like this! I'll need to think about it a little bit since there are two (minor) issues I'd like to consider:

1) this CL will force lib/fs to always be built in all configurations, even in situations where fses aren't really needed (deeply embedded stuff). If it could be made to be conditional if lib/fs is present that might be helpful, though that can be a bit messy in the header space.
2) the switch between using an io_handle_t and a filehandle is a little clumsy. Perhaps it'd make a bit more sense to extend io_handle_t to wrap a file? or even union the two in the FILE so that it's a bit more obvious it's one or the other? I don't know of all of the downstream users of io_handle_t but i suspect there are more than just what's in the tree, so it might be a bit better to test on something ike filehandle being non-null, or store a bool or flag that says which one is active.

For 1 it *is* possoble to test for the presence of another module by testing for WITH_LIB_FS, though it might be a bit nasty ifdef wise.

chiehmin:
@travisg I have push a new commit to fix the issues you mentioned. Please take a look at it to see if it matches your expection. Thanks! 😄

travisg:
Looks pretty good! I might fiddle with it a bit, but I think I can take it as is!

app/tests/stdio_file_tests.c
```diff
@@ -0,0 +1,88 @@
...
+    char buf[1024];
+
+    // test stdout fprintf, fputs
+    fprintf(stdout, "%s", content);
+    fputs(content, stdout);
+
+    // test fwrite
+    wsize = fwrite(content, 1, content_len, stream);
+    if (wsize != content_len) {
+        printf("fwrite size mismatched. wsize = %d, content_len = %d\n", wsize, content_len);
```
travisg, line 31: %d needs to be %zu since it's a size_t. Fails on various architectures.

app/tests/tests.c
```diff
@@ -22,4 +22,7 @@ STATIC_COMMAND("bench", "miscellaneous benchmarks", &benchmarks)
 STATIC_COMMAND("fibo", "threaded fibonacci", &fibo)
 STATIC_COMMAND("spinner", "create a spinning thread", &spinner)
 STATIC_COMMAND("cbuf_tests", "test lib/cbuf", &cbuf_tests)
+#ifdef WITH_LIB_FS
```
travisg, line 25: here and elsewhere, can you replace this with #if instead of #ifdef? There's a convention in the code for using it that way.

lib/libc/stdio.c
```diff
@@ -6,16 +6,26 @@
...
 #include <lk/debug.h>
+#include <lk/err.h>
 #include <stdio.h>
 #include <stdarg.h>
 #include <string.h>
+#include <stdlib.h>
 #include <sys/types.h>
 #include <platform/debug.h>
 
+#ifdef WITH_LIB_FS
```
travisg, line 17: Can probably optimize this a bit to just wrap around the .use_fs line to avoid as much copying of code.

lib/libc/stdio.c
```diff
@@ -24,9 +34,137 @@ FILE __stdio_FILEs[3] = {
...
+        free(stream);
+    }
+    return 0;
+#else
+    return 0;
+#endif // WITH_LIB_FS
+}
+
+size_t fread(void *ptr, size_t size, size_t count, FILE *stream) {
+#ifdef WITH_LIB_FS
```
travisg, line 85: This will probably be a little cleaner to reorder such that the io_read part is the same, but it jsut #ifs in the (stream->use_fs) { } part.

lib/libc/stdio.c
```diff
@@ -24,9 +34,137 @@ FILE __stdio_FILEs[3] = {
...
+    if (NO_ERROR != fs_stat_file(stream->fs_handle.handle, &stat)) {
+        return 1;
+    }
+    return (uint64_t)stream->fs_handle.offset >= stat.size;
+#else
+    return 0;
+#endif // WITH_LIB_FS
+}
+
+static inline int clamp(int val, int min, int max) {
```
travisg, line 119: These types probably should be off_t which is i believe 64bit in 64bit arches.

lib/libc/stdio.c
```diff
@@ -72,10 +229,23 @@ int getchar(void) {
     return getc(stdin);
 }
 
+char *fgets(char *s, int size, FILE *stream) {
+    int c = -1;
+    char *cs = s;
+
```
travisg, line 235: might be worth adding an error check for size < 1, since that should be impossible to fill in a string.

travisg (changes requested):

app/tests/stdio_file_tests.c
```diff
@@ -0,0 +1,88 @@
+#ifdef WITH_LIB_FS
```
travisg, line 1: Since this logically belongs to lib/fs can you move this over into lib/fs/test.c and if possible use the lib/unittest framework?

chiehmin:
@travisg I have resolved all the comments and updated this PR. Please give another review of it. Thanks!

lib/libc/stdio.c
```diff
@@ -42,40 +168,75 @@ int puts(const char *str) {
...
 }
 
-int getc(FILE *fp) {
+int fgetc(FILE *fp) {
     char c;
-    ssize_t ret = io_read(fp->io, &c, sizeof(c));
+    ssize_t ret = fread(&c, /*size=*/1, /*count=*/1, fp);
 
-    return (ret > 0) ? c : ret;
+    return (ret > 0) ? c : -1;
```
chiehmin, line 212: This is a fix for incorrect EOF handling

chiehmin:
@travisg Is there other considerations for this CL?

chiehmin:
@travisg Any update 😃 ?

travisg:
Not yet, haven't had a chance to run it.

travisg:
Just want to comment that I haven't forgotten about this, just have been really busy the last few weeks. Currently on a business trip out of country.

chiehmin:
@travisg Any update for this 😃 ?

travisg:
looks good! apologize for waiting so long, but i'm generally back on LK nowadays and have been going through backlog.

PR #342, opened 2022-08-08
Redundant printf under tcp.c

milisarge:
When tcp_connect function is fired then it prints some time related redundant output then it ruins the console stdout. We can remove extra printf from tcp


travisg (approved):
Looks good, can you re-spin the comment with something a bit more descriptive like:
[tcp] remove extra printf

travisg:
cherry picked and pushed.

PR #343, opened 2022-08-12
[dev][bus][pci] fixed type1_read_byte API bug

eastmoutain:
It should invoke type1_read_byte() rather than type1_read_half().


travisg (approved):
oops!

PR #348, opened 2022-09-03
[libc] add atof() function

milisarge:
I ported atof function from https://github.com/GaloisInc/minlibc repo, i edited it according to code style of LK.

### DIFF
--- a/lib/libc/atof.c
+++ b/lib/libc/atof.c
@@ -1,0 +1,85 @@
+/*
+Copyright (c) 2014 Galois Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions 
+are met:
+
+  * Redistributions of source code must retain the above copyright 
+    notice, this list of conditions and the following disclaimer.
+
+  * Redistributions in binary form must reproduce the above copyright 
+    notice, this list of conditions and the following disclaimer in 
+    the documentation and/or other materials provided with the 
+    distribution.
+
+  * Neither the name of Galois, Inc. nor the names of its contributors 
+    may be used to endorse or promote products derived from this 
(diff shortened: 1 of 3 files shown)


lib/libc/atof.c
```diff
@@ -0,0 +1,59 @@
+/* 
+* Copyright (C) 2014, Galois, Inc.
+* This sotware is distributed under a standard, three-clause BSD license.
+* Please see the file LICENSE, distributed with this software, for specific
+* terms and conditions.
+*/
```
travisg, line 6: Can you replace this with the actual full body of the LICENSE from the repo it came with?

lib/libc/rules.mk
```diff
@@ -10,6 +10,7 @@ MODULE_SRCS += \
 	$(LOCAL_DIR)/abort.c \
 	$(LOCAL_DIR)/atexit.c \
 	$(LOCAL_DIR)/atoi.c \
+	$(LOCAL_DIR)/atof.c \
```
travisg, line 13: You'll want to move this to the MODULE_FLOAT_SRCS section so that appropriate architecture will use floating point compiler switches.

travisg (changes requested):

lib/libc/atof.c
milisarge: i did

lib/libc/rules.mk
milisarge: ah, right.

travisg (approved):
Looks good, can you squash these into a single commit and put a topic that follows the usual LK convention of 

[module] simple one liner

in this case it'd be something like
[libc] add atof() function

travisg:
Perfect, thanks!

PR #351, opened 2022-09-14
Memory mapping issues in ARM64

chintanpandya89:


travisg (approved):
Looks good.

app/tests/mem_tests.c
```diff
@@ -177,7 +177,7 @@ static int mem_test(int argc, const console_cmd_args *argv) {
         }
 
         /* allocate a region to test in */
-        status_t err = vmm_alloc_contiguous(vmm_get_kernel_aspace(), "memtest", len, &ptr, 0, 0, ARCH_MMU_FLAG_UNCACHED);
```
travisg, line 180: Unfortunately this and the above code, in lkboot, rely on uncached mappings in order to properly work.

arch/arm64/mmu.c
```diff
@@ -521,6 +523,35 @@ int arch_mmu_map(arch_aspace_t *aspace, vaddr_t vaddr, paddr_t paddr, uint count
         return NO_ERROR;
 
     int ret;
+
```
travisg, line 526: This routine is a bit problematic due to generally removing the ability to map things as uncached anywhere in practice, except as a hardware register.

Is the problem that since the regular mapping still exists there can exist writeback cache lines that trash the uncached mapping (which seems possible) or that via the large mapping it can prefetch into it? In the latter case that generally seems harmless unless it's secure memory or whatnot, *provided* the memory is cleaned+invalidated (or just invalidated) prior to mapping it here. In that case the extra cached mapping can only at worse end up with a prefetch that ends up with a RO cache line.

kernel/vm/vmm.c
```diff
@@ -370,10 +370,17 @@ status_t vmm_alloc_physical(vmm_aspace_t *aspace, const char *name, size_t size,
 
     /* map all of the pages */
     int err = arch_mmu_map(&aspace->arch_aspace, r->base, paddr, size / PAGE_SIZE, arch_mmu_flags);
-    LTRACEF("arch_mmu_map returns %d\n", err);
+    if (err) {
```
travisg, line 373: These error catching routines here and below look good, can you split them into a separate commit so they can be taken independent of the meat of the change?

travisg (changes requested):
See comments. I'm a bit leery about the fundamental change without talking about it more (can schedule some time next week on chat), but the error handling changes are okay to take if they're split into a separate commit.

travisg:
Closing this out; the two commits landed in different ways.

The `vm map` argument fix is long since in — `kernel/vm/vm.c` reads the argument correctly and checks `arch_mmu_map`'s return.

The mismatched-attribute alias check is a different matter, and it's still an open question rather than a rejected one — I've written it up as #529 along with the parallel proposal in #265. Your report was right, and this tree hit the same class of bug recently under KVM, where Device-mapped shared buffers break on pre-v8.4 ARM hosts; `04c490a3`, `e44365d6` and `941b94ae` moved the virtio virtqueue, 9p PDU buffers and e1000 descriptor rings to cached mappings because of it.

The reason the patch isn't going in as written: the check walks `mmu_initial_mappings` only, so it catches a new mapping that contradicts the boot mapping but not two `vmm_alloc*` mappings that contradict each other, and it's arm64-only. It's also detection rather than prevention — it refuses the alias instead of making a correct mapping reachable. Both of those are worth deciding deliberately, which is what #529 is for. Also, `arch_mmu_map()` is a hot-ish path to put a linear scan of the initial mappings into.

Thanks for the report and the gem5 reproduction.

PR #354, opened 2022-09-22
[dev][power][psci] refactor out of the qemu-virt-arm module

cleverca22:


dev/power/psci/include/dev/power/psci.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
travisg, line 1: Here and others, add the copyright banner.

dev/power/psci/include/dev/power/psci.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include <sys/types.h>
+
+extern int psci_call(ulong arg0, ulong arg1, ulong arg2, ulong arg3);
+extern uint32_t psci_version(void);
+extern int psci_cpu_on(int corenr, ulong entrypoint);
+extern void psci_system_off(void);
+extern void psci_system_reset(void);
+
+#define PSCI_VERSION  0x84000000
```
travisg, line 11: I don't think any of these constants need to be in the header, they can live i the .c file.

travisg (changes requested):
Looks pretty good! just a couple of copyright banner nits and is good to go.

PR #355, opened 2022-09-29
[platform] [riscv-virt] Fix PLIC_PRIORITY macro

heatd:
Old PLIC_PRIORITY macro had an off-by-one bug, causing you to set the priority for the wrong IRQ.


travisg:
Thanks!

PR #357, opened 2022-10-24
 [documentation] update getting started guide

milisarge:
it is mentioned the learn proper toolchain version from GitHub CI workflow config file.


travisg:
Going to close this one out. If the strcspn implementation is needed, will need to be added to the libc rules.mk. Also the documentation is no longer up to date.

PR #358, opened 2022-11-01
[lib][debugcommands] sanity check address is mapped before display/mo…

vamshi51:
…dify

Accessing un-mapped address/address-range results in data abort. Ensure the address and all the pages in the requested range are mapped. Mapping is checked at page size, also adjusts the stop address to the nearest valid address mapped.


travisg (approved):
Looks pretty good to me!

PR #364, opened 2022-12-31
PCI fixes

heatd:
Disables decoding of ioports and memory when touching BAR registers. Also clarifies a function's comments regarding the bottom bits of the BARs.

I didn't touch assign_resource's masking, although it may be a decent idea to do so.


travisg:
LGTM!

PR #365, opened 2023-01-14
Fix CVE-2004-0230

AkanshDivker:
Applied patch for CVE-2004-0230 in tcp_in.c which prevents RST Spoofing Attack (Denial of Service).


travisg:
Oh I'm still thinking about this one, just haven't gotten around to applying it yet.

PR #374, opened 2023-03-10
Add support for building AArch64 with Clang

arichardson:
These changes allow me to build with
`make CC="/usr/bin/clang -target aarch64-unknown-elf" LD=/usr/bin/ld.lld TOOLCHAIN_PREFIX=/usr/bin/ qemu-virt-arm64-test`


arichardson:
Unlike https://github.com/littlekernel/lk/pull/322 this does not add a LLVM= flag to change the default it just allows building if CC/LD are explicitly set to LLVM tools.

arichardson:
ping?

travisg:
These look pretty good, hoping to get a look at it this weekend. Mostly
need to make sure it works well with GCC and see about trying to make it
more generic. There have been multiple attempts to compile with clang over
the years but they're generally fairly limited.

arichardson:
> These look pretty good, hoping to get a look at it this weekend. Mostly need to make sure it works well with GCC and see about trying to make it more generic. There have been multiple attempts to compile with clang over the years but they're generally fairly limited.

Thanks! I did build with GCC as well, and ran the build all script so hopefully it all works. Once these changes are in I'll submit another PR with changes to allow building for RISC-V with clang.

make/macros.mk
```diff
@@ -58,3 +58,17 @@ define MAKECONFIGHEADER
 	echo $3 >> $1.tmp; \
 	$(call TESTANDREPLACEFILE,$1.tmp,$1)
 endef
+
+check_compiler_flag = $(shell $(CC) -c -xc /dev/null -o /dev/null $(1) 2>/dev/null && echo yes || echo no)
+check_warning_supported = $(call check_compiler_flag,-Werror=unknown-warning-option $(1))
```
travisg, line 63: This seems problematic, since gcc doesn't seem to support -Werror=unknown-warning-option. It causes it always to choose that the switch is not supported.

arichardson: Hmm I thought I tested this with GCC but it does look like the docs don't mention it. Will try to update this to only add the flag for clang

travisg: Oh thanks. Been trying to figure out a workaround, but seems that the correct solution is to either only test for clang or to have two paths, based on whether or not the -Werror=unknown-warning-option switch seems to work.

If it's not functional it seems that gcc has some strange behavior with regards to 'no-' switches. If you pass a -Wno- switch that it doesn't recognize it'll accept it unless it gets a compile failure, in which case it then tells you of the weird switch. This seems to be intentional, to allow for backwards compatibility with older versions of the compiler. documented at https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#:~:text=When%20an%20unrecognized%20warning%20option%20is%20requested

travisg: I started to go down the route of just fixing the code that the -Wno- switches were working around, but they're kinda annoying. The big warning failure is in some apis that have the NONNULL attribute set for particular arguments. Inside the function if you test for null anyway (which is a generally good idea, IMO) you get a warning. Kinda annoying, the main reason i had the NONNULL set was to catch obvious null cases at the caller, but not to cause it to elide all checks in the function itself. Really probably means the nonnull attribute is largely a waste of time.

arichardson: Sorry for the delay here - I've pushed a new version that works for both GCC and Clang in my testing.

arichardson: Regarding the __attribute__((nonnull)) that warning probably does make sense since the way that attribute works is that the compiler will elide all NULL checks. Unlike this attribute the Clang _Nonnull/_Nullable annotations only affect compiler warnings and not code generation.

travisg:
LGTM, sorry for the delay!

PR #375, opened 2023-03-10
Clang warning fixes

arichardson:
Fixes various warnings issued when building with Clang.


travisg:
LGTM! Sorry for the late review, been super busy at work and haven't had the energy lately to fiddle with LK on the weekend. But, work is quieting down so have more time/energy.

PR #376, opened 2023-04-05
Candidates for upstream

aaron-odell:
Hi Travis, please review and see which are suitable for upstream. Feel free to provide feedback or reject any that aren't suitable.

### DIFF
--- a/arch/arm/arm-m/systick/systick.c
+++ b/arch/arm/arm-m/systick/systick.c
@@ -31,6 +31,8 @@ static uint32_t tick_rate = 0;
 static uint32_t tick_rate_mhz = 0;
 static lk_time_t tick_interval_ms;
 static lk_bigtime_t tick_interval_us;
+static bool systick_configured;
+static uint32_t reload;
 
 static platform_timer_callback cb;
 static void *cb_args;
@@ -41,9 +43,10 @@ static void arm_cm_systick_set_periodic(lk_time_t period) {
     uint32_t ticks = tick_rate / (1000 / period);
     LTRACEF("ticks %d\n", ticks);
 
-    SysTick->LOAD = (ticks & SysTick_LOAD_RELOAD_Msk) - 1;
+    SysTick->LOAD = reload = (ticks - 1) & SysTick_LOAD_RELOAD_Msk;
     SysTick->VAL = 0;
     SysTick->CTRL = SysTick_CTRL_CLKSOURCE_Msk | SysTick_CTRL_TICKINT_Msk | SysTick_CTRL_ENABLE_Msk;
+    systick_configured = true;
 }
 
 static void arm_cm_systick_cancel_periodic(void) {
@@ -52,9 +55,11 @@ static void arm_cm_systick_cancel_periodic(void) {
 
 /* main systick irq handler */
 void _systick(void) {
-    current_ticks++;
-
     arm_cm_irq_entry();
+
+    if (SysTick->CTRL & SysTick_CTRL_COUNTFLAG_Msk) {
+        current_ticks++;
+    }
 
     bool resched = false;
     if (cb) {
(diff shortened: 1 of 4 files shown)


travisg (approved):

travisg (dismissed):

arch/arm/rules.mk
```diff
@@ -69,6 +69,18 @@ HANDLED_CORE := true
 ENABLE_THUMB := true
 SUBARCH := arm-m
 endif
+ifeq ($(ARM_CPU),cortex-m55)
+GLOBAL_DEFINES += \
+	ARM_CPU_CORTEX_M55=1 \
+	ARM_ISA_ARMv7=1 \
```
travisg, line 75: Will have to probably add proper ARM_ISA_ARMV8 support at some point, but this looks okay, because the compiler has its own intrinsics that CMSIS uses.

top/include/lk/utils.h
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: MIT
+ *
+ * Copyright (C) 2020-2023 Google LLC.
+ */
+
+#pragma once
+
+#define ARRAY_SIZE(_a) (sizeof(_a) / sizeof(*(_a)))
+
+#define KB (1024UL)
```
travisg, line 11: There are some uses of this elsewhere, so will probably need to smush the other defines into this one. Will do.

kernel/thread.c
```diff
@@ -1010,6 +1015,32 @@ void dump_all_threads(void) {
...
+        if (t->magic != THREAD_MAGIC) {
+            dprintf(INFO, "bad magic on thread struct %p, aborting.\n", t);
+            hexdump(t, sizeof(thread_t));
+            break;
+        }
+        if (thread_is_idle(t)) {
+            continue;
+        }
+        // thread specific stats
+        dprintf(INFO, "\t(%s):\n", t->name);
```
travisg, line 1033: Would also probably make sense to dump a per thread version of this inside dump_thread so you see it when you run the `threads` command.

arch/arm/arm-m/systick/systick.c
```diff
@@ -81,38 +86,48 @@ status_t platform_set_periodic_timer(platform_timer_callback callback, void *arg
...
 }
 
+/* Helper to atomically get the global ticks and the current counter value */
+static void systick_get_ticks_val(uint64_t *ticks, uint32_t *val) {
+    spin_lock_saved_state_t state;
+    arch_interrupt_save(&state, SPIN_LOCK_FLAG_INTERRUPTS);
+
+    *val = SysTick->VAL;
+    if (unlikely(SysTick->CTRL & SysTick_CTRL_COUNTFLAG_Msk)) {
+        current_ticks++;
```
travisg, line 96: Isn't there an issue here if this routine is called twice in a row while interrupts are still disabled? It seems like it'd increment the current_ticks twice. Seems like it'd work cleaner if it just returned current_ticks + 1 in this case and still rely on an interrupt getting a chance to fire within the defined interval.

travisg (commented):
Just a few comments that are probably worth looking at. I'm going to take most of these changes into mainline, so a rebase should remove the already taken stuff.

Sorry for the delay! Been pretty pegged at work and haven't had a lot of energy/time to work on LK on the weekend, but things are starting to calm down so I have more time now.

kernel/thread.c
```diff
@@ -519,9 +519,14 @@ void thread_resched(void) {
     if (thread_is_idle(oldthread)) {
         lk_bigtime_t now = current_time_hires();
         thread_stats[cpu].idle_time += now - thread_stats[cpu].last_idle_timestamp;
+    } else {
+        oldthread->stats.total_run_time += current_time_hires() - oldthread->stats.last_run_timestamp;
```
travisg, line 523: Suggestion: since every path through these two if blocks now involves a call to current_time_hires() it'll probably be cleaner to grab a snapshot into a variable `lk_bigtime_t now` before the ifs and just use it inside the loop. Should be the same codegen either way really.

arch/arm/arm-m/systick/systick.c
aaron-odell: The COUNTFLAG bit will only be observed high once per wrap around, so I think this is functionally correct.

From the ARMv7-M Arrchitecture Reference Manual:

> When the counter transitions to zero, it sets the COUNTFLAG status bit to 1. Reading the COUNTFLAG status bit clears it to 0.

The v6m and v8m manuals contain equivalent language. 

Incrementing the counter directly in this case ensures that time remains monotonic even if we are unable to take the systick exception for more than one wraparound period.

kernel/thread.c
aaron-odell: Will do.

arch/arm/arm-m/systick/systick.c
travisg: Ahhhh clear on read. That makes sense. I generally hate clear on read bits which is why I hadn't considered that they had done it but since they had it does seem to clean up this problem. Also since it's implemented all the way back through v6m (lots of cortex-m0s floating around that LK runs on) it works for all cores.

kernel/thread.c
aaron-odell: Done in latest version.

arch/arm/arm-m/systick/systick.c
aaron-odell: Yea the clear on read was surprising to learn about and somewhat annoying, but it works.

I also feel like they missed an opportunity with the placment of the bit. It's in a separate CSR from the actual count value, and we have to consider the possibility of the counter wrapping in between the two separate CSR reads, and do another read if we see the flag set to ensure the value is from after the wrap around. But... the value only occupies the low 24 bits of the readback CSR. They could have just stuck the count flag in one of the upper bits and made it one nice clean atomic read. Oh well.

PR #380, opened 2023-06-07
Allow building RISCV64 and x86_64 with clang and add a Clang CI job

arichardson:


arichardson:
ping :)

engine.mk
```diff
@@ -216,6 +216,14 @@ ARCH_COMPILEFLAGS += -Wno-asm-operand-widths
 endif
 endif
 
+ifeq ($(ARCH),riscv)
```
travisg, line 222: I don't particularly like that this is out in the top level engine.mk, but short of having a second phase arch/rules.mk i'm not sure there's a good other place to put it.

arichardson: I agree this is somewhat ugly. In an earlier version of this patch I added another arch/rules-late.mk include but I feel like that adds quite a bit more complexity. Happy to add it back if you prefer?

travisg: I'll figure something out post-merge. May be time to finally just split the arch/rules.mk into arch/.../toolchain.mk or something, but it would probably be a lot of work.

PR #382, opened 2023-06-21
[arch][arm-m] simplify context switch

fdischner:
The context switch is now always performed inside the PendSV handler, which greatly simplifies the code by reducing all switches to a single path. This should also eliminate any race conditions during the switch.

Because we always enter PendSV for a switch, there is a slight performance penalty in the case of switching from a non-preempted thread to another non-preempted thread (~40 cycles longer on an M4, compared to the previous implementation)

### DIFF
--- a/arch/arm/arm-m/arch.c
+++ b/arch/arm/arm-m/arch.c
@@ -158,7 +158,7 @@ void arm_cm_irq_exit(bool reschedule) {
     target_set_debug_led(1, false);
 
     if (reschedule)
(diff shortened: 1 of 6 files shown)


travisg:
Very interesting, we talked about this before and I'm excited to see how well this works!

travisg:
I pulled it in after some amount of testing. Seems to work here, though going to test a bit more on FPU based cortexes. Thanks!

PR #386, opened 2023-08-30
[arch][arm64] Fix mmu_unmap issue when FEAT_TTL is implemented

shavitmichael:
Precisely set bits [55:22] of the vaddress in bits [43:0] for the vae1is and vaee1is TLBI commands.

On platforms where FEAT_TLL is implemented, bits [47:44] of the command accept a TTL parameter which can optionally be set to hint the translation table level containing the address being invalidated. Implementations aren't architecturally required to perform the invalidation if the hint is incorrect however. Invalidations may therefore fail with the current implementation if the vaddress has bits set in [58:55].

This is notably an issue on ARM fastmodels which doesn't perform the invalidation when the TTL parameter is incorrect.

### DIFF
--- a/arch/arm64/mmu.c
+++ b/arch/arm64/mmu.c
@@ -8,6 +8,7 @@
 
 #include <arch/arm64/mmu.h>
 #include <assert.h>
+#include <lk/bits.h>
 #include <lk/debug.h>
 #include <lk/err.h>
 #include <kernel/vm.h>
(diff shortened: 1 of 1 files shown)


shavitmichael:
Light ping :) @travisg

travisg:
Good catch! Sorry for the delay on this one, I totally missed it in my email spew.

PR #388, opened 2023-10-17
[documentation] LK Memory Management Overview

cradiator:
Add an introduction for LK's memory management system.


travisg:
Looks like a great start, I'll give it a quick read in a sec!

PR #390, opened 2023-10-28
Add CodeQL Workflow for Code Security Analysis

b4yuan:
# Summary

This pull request introduces a CodeQL workflow to enhance the security analysis of this repository.

# What is CodeQL

CodeQL is a static analysis tool that helps identify and mitigate security vulnerabilities. It is primarily intra-function but does provide some support for inter-function analysis. By integrating CodeQL into a GitHub Actions workflow, it can proactively identify and address potential issues before they become security threats.

For more information on CodeQL and how to interpret its results, refer to the GitHub documentation and the CodeQL documentation (https://codeql.github.com/ and https://codeql.github.com/docs/).

# What this PR does

We added a new CodeQL workflow file (.github/workflows/codeql.yml) that
- Runs on every pull request (functionality to run on every push to main branches is included as a comment for convenience).
- Runs daily.
- Excludes queries with a high false positive rate or low-severity findings.
- Does not display results for git submodules, focusing only on our own codebase.

## Validation

To validate the functionality of this workflow, we have run several test scans on the codebase and reviewed the results. The workflow successfully compiles the project, identifies issues, and provides actionable insights while reducing noise by excluding certain queries and third-party code.

## Using the workflow results

If this pull request is merged, the CodeQL workflow will be automatically run on every push to the main branch and on every pull request to the main branch. To view the results of these code scans, follow these steps:
1. Under the repository name, click on the Security tab.
2. In the left sidebar, click Code scanning alerts.

# Is this a good idea?

We are researchers at Purdue University in the USA. We are studying the potential benefits and costs of using CodeQL on open-source repositories of embedded software.

We wrote up a report of our findings so far. The TL;DR is “CodeQL outperforms the other freely-available static analysis tools, with fairly low false positive rates and lots of real defects”. You can read about the report here: https://arxiv.org/abs/2310.00205

# Review of engineering hazards

License: see the license at https://github.com/github/codeql-cli-binaries/blob/main/LICENSE.md: 

```
Here's what you may also do with the Software, but only with an Open Source Codebase and subject to the License Restrictions provisions below:

Perform analysis on the Open Source Codebase.

If the Open Source Codebase is hosted and maintained on GitHub.com, generate CodeQL databases for or during automated analysis, CI, or CD.
```

False positives: We find that around 20% of errors are false positives, but that these FPs are polarized and only a few rules contribute to most FPs. We find that the top rules contributing to FPs are: cpp/uninitialized-local, cpp/missing-check-scanf, cpp/suspicious-pointer-scaling, cpp/unbounded-write, cpp/constant-comparison, and cpp/inconsistent-null-check. Adding a filter to filter out certain rules that contribute to a high FP rate can be done simply in the workflow file.

### DIFF
--- a/.github/workflows/codeql-buildscript.sh
+++ b/.github/workflows/codeql-buildscript.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+make -j$(nproc) lm3s6965evb-test
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -12,8 +12,8 @@
 name: "CodeQL"
 
 on:
-  push:
-    branches: [ "main", "master" ]
+  # push:
+  #   branches: [ "main", "master" ]
   schedule:
     - cron: '0 0 * * *'
   pull_request:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,123 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main", "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main", "master" ]
+  schedule:
+    - cron: '28 21 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners
+    # Consider using larger runners for possible analysis time improvements.
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-20.04' }}
+    timeout-minutes: ${{ (matrix.language == 'swift' && 120) || 360 }}
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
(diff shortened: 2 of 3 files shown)


travisg:
Looks pretty interesting, giving it a check over put probably worth taking.

travisg:
Obvious thing is to try to increase the number of architectures it runs tests on. Is it specifically tied to one toolchain? I see it's using an external arm embedded toolchain.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,127 @@
...
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
+
+    - name: arm-none-eabi-gcc GNU Arm Embedded Toolchain
+      uses: carlosperate/arm-none-eabi-gcc-action@v1.6.0
```
travisg, line 53: This is fine for now, but we should figure out how to extend this to additional arches by using different toolchains, preferably the ones that LK uses for its own builds.

.github/workflows/codeql-buildscript.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+make -j$(nproc) lm3s6965evb-test
```
travisg, line 3: Suggestion: have the target it builds passed in to this script and put the lm3s6965evb in the .yml file.

travisg (commented):
Looks like a good start, though I think running every day is a bit much right now. Comments inline.

travisg:
Thanks for this — CodeQL coverage was worth having, and it's now on master as 2cf992cf with your authorship preserved.

Changes from the PR as posted, mostly to fit how LK actually builds:

- The default CodeQL setup can't build LK, since there's no single "the" build — it's per project, per architecture. The workflow now does a real build using the newos.org toolchain fetch-and-cache pattern from `qemu-boot-tests.yml`, across an arm64/riscv64/x86-64 matrix, so the analysis sees genuinely compiled code for each target.
- Switched from push/PR triggers to a weekly cron (`17 5 * * 1`). At three full cross-builds per run, per-push was more CI time than the signal justified.
- Updated the action versions.

There's since been a follow-on adding `.github/codeql/codeql-config.yml` and `scripts/filter-sarif.py` to filter out the vendored code under `external/`. The scan found and we fixed a real 32-bit block-offset truncation bug class, so this earned its keep.

Closing in favour of the merged workflow.

PR #392, opened 2023-11-10
[make] Remove duplicate compile rules for modes

schultetwin1:
Currently, the compilation rules are mostly duplicated for normal vs float vs ARM source files. This fix deduplicates the compilation logic for these different "mode" of source files by using make's Target-specific variable functionality.

https://www.gnu.org/software/make/manual/make.html#Target_002dspecific

This change was tested by using [`bear`](https://github.com/rizsotto/Bear) to generate a compile_commands.json file. The file was generated both before and after the change and the json diff'd to ensure the generated compile commands were not changed.


schultetwin1:
I tested a project that didn't use any overrides. Closing until I can come up with a fix.

schultetwin1:
Looks like the variable `THUMBCFLAGS` should not be added to `MODULE_ARM_*` compilations. Fixed that!

PR #401, opened 2024-03-01
[lib][heap] Fix array index out-of-range

szsam:
argv[] should have at least 3 elements because argv[2] is accessed in line 345.


travisg (approved):
Nice catch, thanks!

PR #411, opened 2024-06-24
[github-ci] Add uefi_load unittest to github CI

zhangxp1998:


unittest.py
```diff
@@ -0,0 +1,85 @@
+import os
```
travisg, line 1: Can you move this file into scripts/? Usually that's where things like this go, unless it's specific to an application or platform.

travisg:
Looks pretty good, two tweaks:
a) see the file rename comment
b) can you prefix the commit with a [] based topic? in this case would probably be something like [github-ci]. See commits for inspiration.

Thanks!

PR #415, opened 2024-08-08
[vscode] Update workspace file to better respect local style

joshuaseaton:
This change updates lk.code-workspace to specify
* editor.detectIndentation as false, as otherwise that overrides the desired editor.tabSize setting.
* editor.insertSpaces as true, since spaces are preferred over tabs.
* editor.formatOnSave as false for C and C++ files, as - in the absence of a .clang-format file - the clangd extension is likely to default to choices that don't respect the local style.

### DIFF
--- a/lk.code-workspace
+++ b/lk.code-workspace
@@ -1,21 +1,34 @@
 {
-	"folders": [
-		{
-			"path": "."
-		}
-	],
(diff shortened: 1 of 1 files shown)


travisg:
thanks!

PR #417, opened 2024-08-08
[console] List commands alphabetically by name in normal mode

joshuaseaton:
This change updates the `help` console command to list the available commands alphabetically, but only when we're not panicking. This makes `help`'s helping more helpful, which is pretty helpful as far as help goes.


joshuaseaton:
Updated the signature of the comparison function to use `const void*` (which my local GCC v12.2.0 wasn't complaining about)

PR #418, opened 2024-08-11
[scripts] Introduce scripts/fetch-toolchains.py

joshuaseaton:
This change introduces $LK_ROOT/toolchains as an official installation directory for toolchains, as well as a new script for conveniently installing them there. getting_started.md is updated to suggest its use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ scripts/toolpaths.local
 .vscode
 .cache
 compile_commands.json
+toolchain
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -6,25 +6,23 @@
 ```
 mkdir -p lk-work && cd lk-work
 ```
-2- Clone the repo
+2- Clone the repo and change dir to the root
 ```
 git clone https://github.com/littlekernel/lk
-```
-3- Download appropriate toolchain and extract it
-```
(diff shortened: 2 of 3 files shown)


joshuaseaton:
What do you think of updating CI to use this as well?

travisg:
> What do you think of updating CI to use this as well?

I'd like to nak that, because the CI explicitly pulls down what versions it needs, and only exactly what it needs for a specific build. Also it uses the github cache action which saves it from pulling more than once for any given roll.

joshuaseaton:
Fair enough! The needs for CI are certainly different than for devs (e.g., around caching). (This was the classic tension on Fuchsia around sentiments and filed bugs that would amount to "can't you just use fx in automation??")

travisg:
Looks like there's a problem, at least with --list, where it's finding the highest single digit one (ie, 8.2.0, etc) instead of a double digit (ie, 14)

joshuaseaton:
Ah, thanks for catching that. I was doing a naive strcmp on the version string, rather than a lexicographic comp on the actual int tokens. Fixed

travisg (approved):
Looks good, seems to work fine here on ubuntu 24.04 and mac with python 3.12. It may not work right on older pythons, but i'm not particularly worried about that right now unless it becomes a problem.

PR #423, opened 2024-10-10
Implement bulk of UEFI protocol needed by linux kernel

zhangxp1998:


zhangxp1998:
* Added boot service support, implemented common memory functions(allocate/free/get_memory_map), added stub for other functions so that kernel does not crash on call
* Added Runtime service support, only implemented GetVariable/SetVariable, which are called by kernel to get secure boot status.
* Some kernel images might requires relocation, so added relocation support: Load kernel into memory and patch relevant sections according to the relo data.
* Linux Kernel assumes everything is identity mapped, hence anything reachable from the kernel must be identity mapped. This includes the kernel image, kernel stack, memories returned from boot service allocation, etc.

lib/uefi/include/boot_service.h
travisg, line 1: Here and elsewhere in the lib/uefi library, these include files are being added at the top level of the include namespace and being exported to everything.

I'd like it if we could move these headers either into a fully private namespace (ie, not in lib/uefi/include/... which is automatically added to the global namespace) or move them under a directory or two, such as lib/uefi/... so they dont sit at the root of the include heirarchy.

Of particular concern is the 'types.h' file, which is now easily collidable with other stuff.

kernel/include/kernel/vm.h
```diff
@@ -46,6 +46,7 @@ __BEGIN_CDECLS
 
 #define PAGE_ALIGN(x) ALIGN(x, PAGE_SIZE)
 #define IS_PAGE_ALIGNED(x) IS_ALIGNED(x, PAGE_SIZE)
+extern struct list_node arena_list;
```
travisg, line 49: Would be a little cleaner to go ahead and leave this unexported and add a new simple function that returns a pointer to the arena list.

lib/uefi/boot_service_provider.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+vmm_aspace_t *get_address_space() {
+  static vmm_aspace_t *aspace = nullptr;
+  if (aspace == nullptr) {
+    auto err = vmm_create_aspace(&aspace, "linux_kernel", 0);
+    if (err) {
+      printf("Failed to create address space for linux kernel %d\n", err);
+      return nullptr;
+    }
+    vmm_set_active_aspace(aspace);
```
travisg, line 29: This will trash whatever the existing aspace was, which is probably not a huge concern at the moment, but we should perhaps save a copy of the existing one?

lib/uefi/boot_service_provider.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+bool guid_eq(const EfiGuid *a, const EfiGuid *b) {
+  return memcmp(a, b, sizeof(*a)) == 0;
+}
+
+bool guid_eq(const EfiGuid *a, const EfiGuid &b) {
+  return memcmp(a, &b, sizeof(*a)) == 0;
+}
+
+vmm_aspace_t *get_address_space() {
```
travisg, line 21: suggestion: a more descriptive name such as 'set boot aspace' would be clearer what it's actually doing, since it's modifying the existing aspace.

lib/uefi/boot_service_provider.cpp
```diff
@@ -0,0 +1,380 @@
...
+#include "kernel/vm.h"
+#include "lib/dlmalloc.h"
+#include "types.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+EfiStatus unload(EfiHandle handle) { return SUCCESS; }
+
+bool guid_eq(const EfiGuid *a, const EfiGuid *b) {
```
travisg, line 13: here and the rest of this code, if these are local functions they should be marked static or in an anonymous namespace to help the linker inline and GC them.

external/lib/heap/dlmalloc/rules.mk
```diff
@@ -2,6 +2,8 @@ LOCAL_DIR := $(GET_LOCAL_DIR)
 
 MODULE := $(LOCAL_DIR)
 
+MODULE_DEFINES=MSPACES=1
```
travisg, line 5: A comment here as to why this is set would be helpful.

lib/uefi/README.md
```diff
@@ -4,11 +4,13 @@
 make qemu-virt-arm64-test
 ```
 
+Note, this may fail if your system does not have `aarch64-elf-gcc` installed. To address, download from [here](https://newos.org/toolchains/aarch64-elf-14.1.0-Linux-x86_64.tar.xz), unzip, and add the extracted dir to PATH.
```
travisg, line 7: 14.2.0 is the current head.

lib/uefi/boot_service_provider.cpp
```diff
@@ -0,0 +1,380 @@
...
+    printf("Failed to query physical address for memory 0x%p\n", addr);
+    return nullptr;
+  }
+
+  err = arch_mmu_unmap(&aspace->arch_aspace, vaddr, size / PAGE_SIZE);
+  if (err) {
+    printf("Failed to unmap virtual address 0x%lx\n", vaddr);
+    return nullptr;
+  }
+  arch_mmu_map(&aspace->arch_aspace, pa, pa, size / PAGE_SIZE, flags);
```
travisg, line 51: err = ?

lib/uefi/boot_service_provider.cpp
```diff
@@ -0,0 +1,380 @@
...
+}
+
+EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
+                          void **intf) {
+  if (guid_eq(protocol, LOADED_IMAGE_PROTOCOL_GUID)) {
+    printf("handle_protocol(%p, LOADED_IMAGE_PROTOCOL_GUID, %p);\n", handle,
+           intf);
+    auto loaded_image = static_cast<EFI_LOADED_IMAGE_PROTOCOL *>(
+        malloc(sizeof(EFI_LOADED_IMAGE_PROTOCOL)));
+    memset(loaded_image, 0, sizeof(*loaded_image));
```
travisg, line 97: suggestion: since in C++ can do *loaded_image = {};

lib/uefi/boot_service_provider.cpp
```diff
@@ -0,0 +1,380 @@
+
```
travisg, line 1: Here and elsewhere: make sure there's a license on the files.

lib/uefi/switch_stack.S
```diff
@@ -0,0 +1,23 @@
+#include <lk/asm.h>
+
+
+// int call_with_stack(void *stack, int (*fp)(), int param1, int param2);
+FUNCTION(call_with_stack)
+sub     sp,sp,16
```
travisg, line 6: no need to use a separate sub instruction here.

can replace with 
stp    fp, lp, [sp, #-16]!

travisg (changes requested):

travisg:
Looks generally pretty good, just a few changes (such as licenses, etc) and it's at least good enough to not bother other code upstream.

kernel/include/kernel/vm.h
zhangxp1998: Done

lib/uefi/switch_stack.S
```diff
@@ -0,0 +1,23 @@
...
+stp     fp, lr, [sp]
+mov     fp, sp
+
+sub     x0,x0,16
+mov     x6,sp
+str     x6,[x0]
```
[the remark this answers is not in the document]
zhangxp1998: Assembler would not let me. Looks like sp is a special operand which can't be used directly in an str instruction.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done

lib/uefi/include/boot_service.h
zhangxp1998: I digged into the build system a bit, and found https://github.com/littlekernel/lk/blob/master/make/module.mk#L57

```
# add a local include dir to the global include path
GLOBAL_INCLUDES += $(MODULE_SRCDIR)/include
```

All modules were adding its MODULE_SRCDIR/include to global namespace, so it became the default. I have moved all UEFI headers to lib/uefi/local/include, to avoid polluting the global namespace.

external/lib/heap/dlmalloc/rules.mk
zhangxp1998: Done

lib/uefi/README.md
zhangxp1998: Done

travisg (approved):
LGTM!

PR #427, opened 2025-01-02
[arch][arm] Enable distributor with ARE, group 1

gilbert0225:
Refering Zircon's 'gic_init' function, enable ARE and Group 1 added.
For arm_generic_timer configuration, 'arm64_get_boot_el' and 'arm64_is_kernel_in_hyp_mode' added to make decision for the hypervisor/virtual timer irq.

### DIFF
--- a/arch/arm64/arch.c
+++ b/arch/arm64/arch.c
@@ -27,6 +27,13 @@ static spin_lock_t arm_boot_cpu_lock = 1;
 static volatile int secondaries_to_init = 0;
 #endif
 
+/* Defined in start.S. */
+extern uint64_t arm64_boot_el;
+
+uint64_t arm64_get_boot_el(void) {
+    return arm64_boot_el >> 2;
+}
+
 static void arm64_cpu_early_init(void) {
     /* set the vector base */
     ARM64_WRITE_SYSREG(VBAR_EL1, (uint64_t)&arm64_exception_base);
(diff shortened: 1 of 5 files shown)


arch/arm64/arch.c
```diff
@@ -27,6 +27,13 @@ static spin_lock_t arm_boot_cpu_lock = 1;
 static volatile int secondaries_to_init = 0;
 #endif
 
+/* Defined in start.S. */
+extern uint64_t arm64_boot_el;
+
+uint64_t arm64_get_boot_el() {
```
travisg, line 33: C nit: needs (void) instead of () for language reasons.

arch/arm64/include/arch/arm64.h
```diff
@@ -82,5 +84,16 @@ void arm64_local_invalidate_cache_all(void);
 void arm64_local_clean_invalidate_cache_all(void);
 void arm64_local_clean_cache_all(void);
 
+
+/* Current Exception Level values, as contained in CurrentEL */
+#define CurrentEL_EL1		(1 << 2)
+#define CurrentEL_EL2		(2 << 2)
+
+static inline bool arm64_is_kernel_in_hyp_mode(void)
```
travisg, line 92: tiny nit: can put { on the same line and use space indents for line 94?

travisg (changes requested):
Only a few little nits but otherwise great, thanks!

travisg:
Note this still doesn't work on qemu arm64 with gic version set to 3, but this was definitely an oversight.

arch/arm64/arch.c
gilbert0225: Resolved

arch/arm64/include/arch/arm64.h
gilbert0225: Resolved

travisg:
Thanks!

PR #428, opened 2025-01-03
[kernel][port] add a missing THREAD_UNLOCK call in port_group_remove()

namyoonw:
port_group_remove() used to return without unlocking a thread if no matching read port is found, which is not correct. This patch adds a call to THREAD_UNLOCK before it returns ERR_BAD_HANDLE.


travisg (approved):
Yep, you're right! Thanks for the fix!

PR #429, opened 2025-01-06
Add block IO protocol to UEFI loader

zhangxp1998:
Changes made to other parts of LK:

1. Added a get_fdt() function to platform.c and platform.h, each platform should provide its own implementation 
2. Added a iter_block_device() function to bio.c, UEFI uses this function to provide block IO protocol to UEFI application


lib/bio/include/lib/bio.h
```diff
@@ -130,4 +130,18 @@ enum bio_ioctl_num {
     BIO_IOCTL_IS_MAPPED,    /* if supported, returns whether or not the device is memory mapped. */
 };
 
-__END_CDECLS
\ No newline at end of file
+void bio_iter_devices(bool (*callback)(void *, bdev_t *), void *cookie);
```
travisg, line 133: Can you put a comment here that it is called back with the internal bio mutex held, so that callers dont try to do other bio routines inside the callback?

travisg: Similarly can you add a comment that the bool of the callback tells it to continue calling back or not?

lib/uefi/configuration_table.cpp
```diff
@@ -17,15 +17,24 @@
 
 #include "configuration_table.h"
 #include "boot_service_provider.h"
+#include "libfdt.h"
+#include "platform.h"
 #include "system_table.h"
 #include <string.h>
 
 void setup_configuration_table(EfiSystemTable *table) {
+  const void *fdt = get_fdt();
```
travisg, line 26: Would be pretty wise to null check this.

platform/include/platform.h
```diff
@@ -82,5 +82,8 @@ void platform_init(void);
 /* called by the arch init code to get the platform to set up any mmu mappings it may need */
 void platform_init_mmu_mappings(void);
 
+/* Called by LK to get the device tree */
+const void *get_fdt(void);
```
travisg, line 86: Can you add a default weak reference version of this in platform/init.c (see the other __WEAK) functions, so that there's a default version that returns a null?

travisg (changes requested):
Looks pretty good with a few little nits.

lib/bio/include/lib/bio.h
zhangxp1998: Done

lib/uefi/configuration_table.cpp
zhangxp1998: Done

platform/include/platform.h
zhangxp1998: Done

PR #432, opened 2025-01-10
[lib][uefi] fix a few warnings and a little code tidying

travisg:
GCC 14 is quite picky about warnings, probably more so than clang.

-Fix a bunch of printf warnings. Pointers should be printed with %p. -Move some stuff into an anonymous namespace.
-Worked around GCC really not liking reinterpret_casting from one function pointer type to another. Fiddled with it a bit and eventually settled on casting the function pointer to const void * and passing it through.

### DIFF
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -421,10 +421,10 @@ EfiStatus read_blocks(EfiBlockIoProtocol *self, uint32_t media_id, uint64_t lba,
     return END_OF_MEDIA;
   }
 
-  const auto bytes_read =
+  const size_t bytes_read =
       call_with_stack(interface->io_stack, bio_read_block, dev, buffer, lba,
                       buffer_size / dev->block_size);
-  if (bytes_read != static_cast<ssize_t>(buffer_size)) {
+  if (bytes_read != buffer_size) {
     printf("Failed to read %ld bytes from %s\n", buffer_size, dev->name);
     return DEVICE_ERROR;
   }
@@ -449,7 +449,7 @@ EfiStatus open_block_device(EfiHandle handle, void **intf) {
     vmm_alloc(vmm_get_kernel_aspace(), "uefi_io_stack", kIoStackSize, &io_stack,
               PAGE_SIZE_SHIFT, 0, 0);
   }
-  printf("%s(%s)\n", __FUNCTION__, handle);
+  printf("%s(%p)\n", __FUNCTION__, handle);
   const auto interface = reinterpret_cast<EfiBlockIoInterface *>(
       mspace_malloc(get_mspace(), sizeof(EfiBlockIoInterface)));
   memset(interface, 0, sizeof(EfiBlockIoInterface));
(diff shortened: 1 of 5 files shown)


lib/uefi/boot_service_provider.cpp
```diff
@@ -424,7 +424,7 @@ EfiStatus read_blocks(EfiBlockIoProtocol *self, uint32_t media_id, uint64_t lba,
   const auto bytes_read =
       call_with_stack(interface->io_stack, bio_read_block, dev, buffer, lba,
                       buffer_size / dev->block_size);
-  if (bytes_read != static_cast<ssize_t>(buffer_size)) {
+  if (bytes_read != buffer_size) {
```
zhangxp1998, line 427: clang issues a warning when comparing signed integers with unsigned ones. Which is why there's a static cast in the first place.

lib/uefi/boot_service_provider.cpp
```diff
@@ -692,7 +692,7 @@ EfiStatus locate_handle_buffer(EfiLocateHandleSearchType search_type,
     }
     return SUCCESS;
   }
-  printf("%s(0x%x, (0x%x 0x%x 0x%x 0x%llx), search_key=0x%lx)\n", __FUNCTION__,
+  printf("%s(0x%x, (0x%x 0x%x 0x%x 0x%llx), search_key=%p)\n", __FUNCTION__,
```
zhangxp1998, line 695: Different compilers seem to have different implementation for %p, some include 0x in the front some don't. Is there a way to enforce a consistent formatting for pointers?

travisg: It's not implemented in the compiler actually. In LK's case it's implemented in the libc. The spec says that it will put a 0x in front of it if it's not zero, but if it's zero it just shows that. Kinda weird, but I get used to it.

Similar to %#x (instead of 0x%x). The #x is the nicer, shorter way of doing it, but it only puts a 0x if it's nonzero.

lib/uefi/boot_service_provider.cpp
travisg: Hmm, in this case since it was an auto it was deducing a unsigned version. As a general rule I wouldn't overuse auto for anything but complicated types that are hard to type out, for exactly this sort of reason.

If you haven't tried it before, building with -Werror is a good idea.

lib/uefi/boot_service_provider.cpp
zhangxp1998: ack

lib/uefi/boot_service_provider.cpp
travisg: Yeah, I double checked this. call_with_stack returns a size_t, so in this case it's an unsigned vs unsigned. I can change the type so it doesn't use an auto to make sure though.

travisg: pushed a new version that removes the auto and makes sure it's a size_t (though it will be because call_with_stack returns type size_t).

WDYT?

zhangxp1998: SGTM

PR #438, opened 2025-02-27
[lib][miniheap] fix: modify the assert condition in miniheap

vivekj-samsung:
For LK_DEBUGLEVEL > 1
> alloc_struct_begin_size: 24
> free_heap_chunk_size: 24

size: max(alloc_struct_begin_size, free_heap_chunk_size)

But when freeing the chunk, allocated size is expected to be greater than size of free_heap_chunk struct.
It contradicts its own code.
So add >= instead of > to maintain the integrity between allocation and freeing of memory chunk.

### DIFF
--- a/lib/heap/miniheap/miniheap.c
+++ b/lib/heap/miniheap/miniheap.c
@@ -149,7 +149,7 @@ try_merge:
 
 static struct free_heap_chunk *heap_create_free_chunk(void *ptr, size_t len, bool allow_debug) {
     DEBUG_ASSERT((len % sizeof(void *)) == 0); // size must be aligned on pointer boundary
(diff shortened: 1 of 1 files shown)


vivekj-samsung:
Hi @travisg 

Could you please review this pull request when you have time?

Thank you!

travisg:
Sure thing, looking at it now.

travisg:
Interesting, it looks like this does indeed fix a rare case with an allocation of zero length being freed later. I'm not sure why I've never seen this before, though I think most 64bit builds are using cmpctmalloc instead of miniheap so maybe that's why. LGTM!

vivekj-samsung:
Thanks!

PR #439, opened 2025-03-13
[lib][uefi] initialize uefi system table with zeroes

dimorinny:
Mediatek hit `table->number_of_table_entries` isn't initialized

### DIFF
--- a/lib/uefi/uefi.cpp
+++ b/lib/uefi/uefi.cpp
@@ -104,6 +104,7 @@ int load_sections_and_execute(bdev_t *dev,
   printf("Entry function located at %p\n", entry);
 
   EfiSystemTable &table = *static_cast<EfiSystemTable *>(alloc_page(PAGE_SIZE));
+  memset(&table, 0, sizeof(EfiSystemTable));
   EfiBootService boot_service{};
   EfiRuntimeService runtime_service{};
   fill(&runtime_service, 0);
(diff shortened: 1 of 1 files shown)


lib/uefi/uefi.cpp
```diff
@@ -106,6 +106,7 @@ int load_sections_and_execute(bdev_t *dev,
   EfiSystemTable &table = *static_cast<EfiSystemTable *>(alloc_page(PAGE_SIZE));
   EfiBootService boot_service{};
   EfiRuntimeService runtime_service{};
+  fill(&table, 0);
```
zhangxp1998, line 109: use memset, fill will fill the table with increasing byte pattern.

dimorinny: Done, thank you. Just for my interest, what is the purpose of filling `EfiBootService` and `EfiRuntimeService` with increasing byte pattern?

zhangxp1998: The runtime and bootservice table are huge, and not all function pointers in these tables are implemented. When GBL crashes due to calling into unimplemented protocols, an increasing byte pattern would tell us which protocol needs to be implemented.

e.g. if GBL crashes at address 0xE0E1E2E3E4 I know that it crashes by calling into a function pointer at offset 0xE0. Versus if you zero initialize everything, all I can see is a null pointer exception, and I would have no idea which field need to be implemented.

dimorinny: Thanks for sharing, I like the idea. Updated the CL, so also covered `table.configuration_table` with memset.

travisg: Looks good to me, should I take it @zhangxp1998 ?

zhangxp1998 (approved):

PR #442, opened 2025-04-09
atol handles overflow and accepts leading whitespace and '+'

HBehrens:
Make `atol()` more robust and compatible with gclib's implementation that

1. accepts leading whitespace
2. accepts leading '+' character
3. handles overflow

Similar adjustments should be made to `atoul` and `atoull`.
I wasn't able to run this project's test suite and since my own (closed source)
project only covers `atol` (see tests in comment), I did not want to
touch anything but this function.


travisg (approved):
looks good to me. thanks!

PR #444, opened 2025-05-04
Add FVP Base support

toor1245:
Add initial support of FVP Base platform, supported peripherals:

* Arm Virtual Timer
* GICv2
* UART

uart0 log: [log.txt](https://github.com/user-attachments/files/20028953/log.txt)

### DIFF
--- a/docs/fvp-base/how_to_run.md
+++ b/docs/fvp-base/how_to_run.md
@@ -1,0 +1,80 @@
+# Build and run Fixed Virtual Platform (FVP)
+
+This guide introduces how to build and run little kernel on FVP Base platform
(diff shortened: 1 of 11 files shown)


toor1245:
@travisg, I see failed tests, I added a new platform, it is unlikely that I affected something to qemu platform, could you please take a look?

travisg:
Oh thanks! Yeah the faiures are on the github runner side where for some reason it sometimes fails to apt-get the qemu runtime. Dunno whats up there.

Will take a look soon!

travisg:
Very cool! Thanks for the input, this will work well.

PR #445, opened 2025-05-28
[lib][uefi] Make load_pe_file extern linkage

zhangxp1998:
This function was previously put in anon namespace, making it unavailable for other source files to consume. Since we made this function an API for consumption, move outside anon namespace so that linkers can find it.


zhangxp1998:
(Merged this because it's a small change and really only affects UEFI code

travisg:
Sounds good, thanks for the CI fix. I've been meaning to look into it, but it was really unclear why it's failing. If this fixes it, yay.

PR #446, opened 2025-05-29
Allow complete platform customization for UEFI

zhangxp1998:
Added platform_setup_system_table, where platform/vendor can make arbitrary changes to UEFI's system table, overriding any of the UEFI protocols with custom one.

`platform_setup_system_table` has a default implementation which is weakly linked, this allows platforms/vendors to provide their own implementation of protocols in a separate directory. As a result, all platform/vendor customizations to UEFI can be done without touching lib/uefi, making future merges easier.

To support this, we need to make UEFI protocol definitions available outside of lib/uefi, which is done by moving headers from local/include to include/uefi . All UEFI headers are put under a `uefi` directory to avoid name collision in global include namespace.


travisg (approved):
LGTM

PR #451, opened 2025-07-10
Reset errno values where read

schultetwin1:
According to `man 3 errno`:

```
The value of errno is never set to zero by any system
call or library function.
```

So we must set it to 0 ourselves before each `read` and `write` call to ensure we don't get stuck in an error loop.

### DIFF
--- a/tools/bootimage.c
+++ b/tools/bootimage.c
@@ -96,6 +96,7 @@ static int writex(int fd, void *data, size_t len) {
     int r;
     char *x = data;
     while (len > 0) {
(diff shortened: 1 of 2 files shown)


swetland:
This change is not harmful, but not necessary.
errno is only being checked where read() or write() return an error condition, in which case errno has been set.
```
 On  error,  -1  is returned, and errno is set to indicate the error.
```

travisg:
Oh hah i just now noticed this was in the tools code not libc. Oh well, harmless.

PR #452, opened 2025-07-16
Support os_loading protocol and async io protocol

zhangxp1998:


lib/uefi/blockio2_protocols.cpp
```diff
@@ -0,0 +1,123 @@
+
```
travisg, line 1: Here and below, make sure you add a copyright header to each file.

lib/uefi/thread_utils.h
```diff
@@ -0,0 +1,28 @@
...
+template <typename Func>
+int functor_thread_trampoline(void *f) {
+  auto callable = reinterpret_cast<Func *>(f);
+  int ret = (*callable)();
+  delete callable;
+  return ret;
+}
+
+template <typename Func>
+thread_t *thread_create(const char *name, Func f, int priority,
```
travisg, line 20: might be less confusing to name this the same as the underlying C version.

lib/uefi/blockio2_protocols.cpp
```diff
@@ -0,0 +1,123 @@
...
+        signal_event(token->event);
+        return 0;
+      },
+      get_current_thread()->priority, kIoStackSize);
+  if (thread == nullptr) {
+    printf("Failed to create thread for IO read\n");
+    return DEVICE_ERROR;
+  }
+
+  auto err = thread_resume(thread);
```
travisg, line 70: If the thread isn't detached, then the remants of the thread will stick around forever until something calls thread_join on it. If the intention is for someone to poll for transaction_status to be SUCCESS, then you should detach it here.

travisg (changes requested):
i didn't look super closely, but this seems to work with some changes.

lib/uefi/thread_utils.h
zhangxp1998: The underlying C version is also called `thread_create` , this is what the C++ version is named. I don't see what needs to change.

travisg: Oh I thought it was just a bit confusing to name a new C++ overriding function that same name as the underlying C one.

zhangxp1998: Done, renamed to thread_create_functor

travisg (approved):
LGTM

PR #453, opened 2025-07-24
[lib][uefi] Implement EFI_TIMESTAMP_PROTOCOL

silverneko:
Reads the ARM64-specific counter registers.
This only works on ARM64 platform obviously, which is the only platform we support for UEFI at the moment.

### DIFF
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -239,10 +239,9 @@ EfiStatus open_protocol(EfiHandle handle, const EfiGuid *protocol, void **intf,
            __FUNCTION__, handle, agent_handle, controller_handle, attr);
     return SUCCESS;
   } else if (guid_eq(protocol, EFI_DEVICE_PATH_PROTOCOL_GUID)) {
-    printf(
-        "%s(EFI_DEVICE_PATH_PROTOCOL_GUID, handle=%p, agent_handle=%p, "
-        "controller_handle=%p, attr=0x%x)\n",
-        __FUNCTION__, handle, agent_handle, controller_handle, attr);
+    printf("%s(EFI_DEVICE_PATH_PROTOCOL_GUID, handle=%p, agent_handle=%p, "
+           "controller_handle=%p, attr=0x%x)\n",
+           __FUNCTION__, handle, agent_handle, controller_handle, attr);
     return UNSUPPORTED;
   } else if (guid_eq(protocol, EFI_BLOCK_IO_PROTOCOL_GUID)) {
     printf("%s(EFI_BLOCK_IO_PROTOCOL_GUID, handle=%p, agent_handle=%p, "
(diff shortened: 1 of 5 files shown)


lib/uefi/uefi_platform.cpp
```diff
@@ -104,3 +105,16 @@ __WEAK EfiStatus platform_setup_system_table(EfiSystemTable *table) {
...
+
+__WEAK uint64_t get_timestamp() {
+  return ARM64_READ_SYSREG(cntpct_el0);
+}
+
+__WEAK EfiStatus get_timestamp_properties(EfiTimestampProperties *properties) {
+  if (properties == nullptr) {
+    return INVALID_PARAMETER;
+  }
+  properties->frequency = static_cast<uint32_t>(ARM64_READ_SYSREG(cntfrq_el0));
```
silverneko, line 117: I'm doing this weird converting to and from uint64 <-> uint32 because according to spec, only the lower 32 bits maps to the clock frequency value, so I truncate the register value by down-casting to uint32_t, and then immediately cast it back to uint64_t

zhangxp1998: Maybe we can use `& 0xFFFFFFFF` ? This communicates the intent in an easier way.

lib/uefi/boot_service_provider.cpp
```diff
@@ -237,6 +237,19 @@ EfiStatus open_protocol(EfiHandle handle, const EfiGuid *protocol, void **intf,
     config->select_device_trees = select_device_trees;
     *intf = reinterpret_cast<void *>(config);
     return SUCCESS;
+  } else if (guid_eq(protocol, EFI_TIMESTAMP_PROTOCOL_GUID)) {
+    printf("%s(EFI_TIMESTAMP_PROTOCOL_GUID, handle=%p, agent_handle=%p, "
+           "controller_handle=%p, attr=0x%x)\n",
+           __FUNCTION__, handle, agent_handle, controller_handle, attr);
+    EfiTimestampProtocol *ts = nullptr;
+    allocate_pool(BOOT_SERVICES_DATA, sizeof(EfiTimestampProtocol), reinterpret_cast<void**>(&ts));
```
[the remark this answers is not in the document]
silverneko: when would I need to use `allocate_pool` pool then? what makes say for example `EfiDtFixupProtocol` special?

lib/uefi/uefi_platform.cpp
silverneko: > Maybe we can use `& 0xFFFFFFFF` ? This communicates the intent in an easier way.

I don't have preferences here. Suggestion applied.

silverneko: > Should we call `get_timestamp` here instead of `ARM64_READ_SYSREG` ?

I don't understand the question? We are reading the `cntfrq` register here to get the frequency of the timer. 
`get_timestamp` OTOH reads the `cntpct` reg.

lib/uefi/boot_service_provider.cpp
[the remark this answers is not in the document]
zhangxp1998: `EfiDtFixupProtocol` isn't special, I should have just used uefi_malloc or even malloc. `allocate_pool` takes a `pool_type` argument, but current implementation just ignores this argument and calls `mspace_malloc` . `uefi_malloc` also just calls `mspace_malloc`, so these two are identical, they allocate identity mapped memory from pre-allocated heap memory.

PR #454, opened 2025-07-31
[lib][uefi] Free memory used by UEFI app after app returns

zhangxp1998:


lib/uefi/events.cpp
```diff
@@ -86,47 +98,39 @@ EfiStatus wait_for_event(size_t num_events, EfiEvent *event, size_t *index) {
...
     // If this event is signaled on a different thread,  defer
     // calling callbacks until the next check_event call. As UEFI apps
     // are single threaded, we don't want to call event callbacks from another
     // thread
-    if (event->creator_thread != get_current_thread()) {
+    if (event->creator_thread != get_current_thread() || arch_ints_disabled()) {
       LTRACEF(
           "Event %p of type 0x%x is signaled from thread %s, defer notify_fn "
-          "because event is created on thread %s\n",
+          "because event is created on thread %s(or interrupt is disabled)\n",
```
silverneko, line 117: how about dropping the parenthesis? Just say "because event is created on other thread or interrupt is disabled"

lib/uefi/events.cpp
```diff
@@ -86,47 +98,39 @@ EfiStatus wait_for_event(size_t num_events, EfiEvent *event, size_t *index) {
...
-        if (!success) {
-          thread_yield();
-        }
-      }
       return SUCCESS;
     }
     event->notify_fn(event, event->notify_ctx);
     event->callback_called = true;
+    // this is only possible in non-interrupt context, as this requires mutexes.
+    delete_if_in_list(event);
```
silverneko, line 125: This has a lot in common with `process_pending_events()`, how about making this a helper function?

lib/uefi/events.cpp
```diff
@@ -30,30 +31,39 @@
...
-lk_time_t backoff_wait_time = 1;
+struct list_node pending_events = LIST_INITIAL_VALUE(pending_events);
+
+void process_pending_events() {
+  AutoLock al{&event_list_mutex};
+  EfiEventImpl *ev = nullptr;
+  list_for_every_entry(&pending_events, ev, EfiEventImpl, node) {
+    if (ev->ready()) {
+      if (!ev->callback_called && ev->notify_fn != nullptr) {
+        LTRACEF("Triggering event %p callback %p %p on thread %s\n", ev,
```
silverneko, line 47: `"Triggering event %p callback %p ctx %p on thread %s\n"`

lib/uefi/uefi.cpp
```diff
@@ -124,17 +124,20 @@ int load_sections_and_execute(bdev_t *dev,
...
   }
 
-  constexpr size_t kStackSize = 8 * 1024ul * 1024;
+  constexpr size_t kStackSize = 1 * 1024ul * 1024;
   auto stack = reinterpret_cast<char *>(alloc_page(kStackSize, 23));
   memset(stack, 0, kStackSize);
   printf("Calling kernel with stack [%p, %p]\n", stack, stack + kStackSize - 1);
-  return call_with_stack(stack + kStackSize, entry, image_base, &table);
+  auto ret = call_with_stack(stack + kStackSize, entry, image_base, &table);
+  reset_heap();
```
silverneko, line 139: How about using a RAII-style ScopeGuard to call reset_heap()? This ensures that we don't forget to call it before return? For example, the return statement at L131 is missing a call to reset_heap.

lib/uefi/uefi.cpp
```diff
@@ -124,17 +124,20 @@ int load_sections_and_execute(bdev_t *dev,
...
+  setup_heap();
   auto status = platform_setup_system_table(&table);
   if (status != SUCCESS) {
     printf("platform_setup_system_table failed: %lu\n", status);
     return -static_cast<int>(status);
   }
 
-  constexpr size_t kStackSize = 8 * 1024ul * 1024;
+  constexpr size_t kStackSize = 1 * 1024ul * 1024;
   auto stack = reinterpret_cast<char *>(alloc_page(kStackSize, 23));
```
silverneko, line 135: shouldn't we free this memory region after call_with_stack returned?

lib/uefi/memory_protocols.cpp
```diff
@@ -45,15 +48,36 @@ mspace create_mspace_with_base_limit(void *base, size_t capacity, int locked) {
...
+EfiStatus free_pages(void *memory, size_t pages) {
+  return ::free_pages(reinterpret_cast<EfiPhysicalAddr>(memory), pages);
+}
+
 } // namespace
 
+void setup_heap() {
+  if (space == nullptr) {
+    space = create_mspace_with_base_limit(get_heap(), kHeapSize, 1);
+  }
```
silverneko, line 71: just call `get_mspace()` ?

lib/uefi/memory_protocols.cpp
```diff
@@ -45,15 +48,36 @@ mspace create_mspace_with_base_limit(void *base, size_t capacity, int locked) {
...
   return space;
 }
 
 void restore_aspace() { vmm_set_active_aspace(old_aspace); }
 
+EfiStatus free_pages(void *memory, size_t pages) {
+  return ::free_pages(reinterpret_cast<EfiPhysicalAddr>(memory), pages);
```
silverneko, line 63: this is only called once, why not just inline it?

lib/uefi/events.cpp
zhangxp1998: Done

lib/uefi/events.cpp
```diff
@@ -30,30 +31,39 @@
...
+                ev->notify_fn, ev->notify_ctx, get_current_thread()->name);
+        backoff_wait_time = 0;
+        ev->notify_fn(ev, ev->notify_ctx);
+        ev->callback_called = true;
+      }
+      if (ev->node.prev != nullptr && ev->node.next != nullptr) {
+        list_delete(&ev->node);
+        return;
```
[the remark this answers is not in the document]
zhangxp1998: Because we have modified the list, continue iteration would crash.

lib/uefi/events.cpp
zhangxp1998: Done

lib/uefi/uefi.cpp
zhangxp1998: Done, using DEFER macro(we don't have unique_ptr in LK codebase

lib/uefi/events.cpp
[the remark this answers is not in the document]
silverneko: I think you want to just use the "safe" variant list macros `list_for_every_entry_safe`
https://github.com/littlekernel/lk/blob/master/top/include/lk/list.h#L231

lib/uefi/events.cpp
silverneko: How about change `delete_if_in_list` to not hold the lock, but instead require the caller of `delete_if_in_list()` to acquire the lock?

lib/uefi/memory_protocols.cpp
silverneko: wait, why not just change the prototype of `free_pages` from `free_pages(EfiPhysicalAddr memory, size_t pages)` to `free_pages(void* memory, size_t pages)` ? This way any pointer type can be implicitly cast to `void*`.

I just found it weird that we are casting back and forth multiple times to and from `void*`. `free_pages (overload)` cast `void*` to `EfiPhysicalAddr`, and `free_pages (original)` cast the `EfiPhysicalAddr` back to `void *` immediately... then why not just use `void* all the way?

The only place `EfiPhysicalAddr` is relevant is when `free_pages` is called from UEFI app. We shall only explicitly cast from `EfiPhysicalAddr` to `void*` when calling `free_pages` from UEFI.

lib/uefi/events.cpp
```diff
@@ -30,30 +31,49 @@
...
+  EfiEventImpl *ev = nullptr;
+  EfiEventImpl *to_delete = nullptr;
+  list_for_every_entry(&pending_events, ev, EfiEventImpl, node) {
+    if (to_delete != nullptr) {
+      list_delete(&to_delete->node);
+      to_delete = nullptr;
+    }
+    if (ev->ready()) {
+      if (!ev->callback_called && ev->notify_fn != nullptr) {
+        LTRACEF("Triggering event %p callback %p %p on thread %s\n", ev,
```
silverneko, line 52:
```suggestion
        LTRACEF("Triggering event %p callback %p ctx %p on thread %s\n", ev,
```

lib/uefi/events.cpp
zhangxp1998: It would cause code duplication in call sites which are not already holding the lock

lib/uefi/uefi.cpp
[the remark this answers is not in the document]
zhangxp1998: Done

lib/uefi/events.cpp
```diff
@@ -86,47 +115,40 @@ EfiStatus wait_for_event(size_t num_events, EfiEvent *event, size_t *index) {
...
-              events[i++] = event;
-              success = true;
-            });
-        if (!success) {
-          thread_yield();
-        }
-      }
       return SUCCESS;
     }
     event->notify_fn(event, event->notify_ctx);
```
silverneko, line 140: should we reset the backoff timer here too?

should we just call `invoke_callback()` here?

lib/uefi/events.cpp
```diff
@@ -30,30 +31,56 @@
...
-    f(arr, events);
+
+lk_time_t backoff_wait_time = 1;
+
+Mutex event_list_mutex;
+
+struct list_node pending_events = LIST_INITIAL_VALUE(pending_events);
+
+bool invoke_callback(EfiEventImpl *ev) {
+  DEBUG_ASSERT(ev->ready());
```
silverneko, line 42: suggestion: Assert that the current thread is the main thread

lib/uefi/events.cpp
silverneko: which I think is a net gain in readability and usability? I think the code would be more readable if the lock is acquired explicitly by the caller vs. acquired implicitly by the callee. If the "acquire lock" step is done explicitly by the caller, the caller would be forced to do lock management and acknowledge of potential deadlock. 

moving the lock out of `delete_if_in_list()` also improves usability by allowing it to be combined with other actions in a single critical section.

lib/uefi/events.cpp
```diff
@@ -193,13 +205,20 @@ EfiStatus check_event(EfiEvent event) {
     return INVALID_PARAMETER;
   }
   if (event->ready()) {
+    invoke_callback(event);
```
silverneko, line 208: I suspect this is a no-op, because if the current is ready, it's callbakc would've been executed in `process_pending_events`, thus the `event->callback_called` must be true, thus this would do nothing, am I right?

lib/uefi/events.cpp
```diff
@@ -30,30 +31,56 @@
...
-struct {
- public:
-  template <typename Func>
-  void read_access(Func f) const {
-    AutoLock a(&m);
-    f(arr, events);
+
+lk_time_t backoff_wait_time = 1;
+
+Mutex event_list_mutex;
```
silverneko, line 37: I just noticed that the mutex is only ever accessed by the main UEFI thread. Why would we need this mutex then?

lib/uefi/events.cpp
```diff
@@ -30,30 +31,56 @@
...
+}
 
-lk_time_t backoff_wait_time = 1;
+void process_pending_events() {
+  AutoLock al{&event_list_mutex};
+  EfiEventImpl *ev = nullptr;
+  EfiEventImpl *tmp_ev = nullptr;
+  list_for_every_entry_safe(&pending_events, ev, tmp_ev, EfiEventImpl, node) {
+    invoke_callback_if_ready(ev);
+    if (ev->callback_called) {
```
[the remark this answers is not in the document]
zhangxp1998: We do want to remove the callback if callback was already called before runing this block of code

lib/uefi/events.cpp
zhangxp1998: Done

lib/uefi/events.cpp
zhangxp1998: Yea, unless event was signaled between process_pending_events() and this line, but we can probably drop it

lib/uefi/events.cpp
zhangxp1998: We have threading based async I/O which may invoke events API on other threads. In the future we might also add other threaded feature, so I made events thread safe.

lib/uefi/events.cpp
```diff
@@ -30,30 +31,55 @@
...
+      // we can't call |delete_if_in_list| here because mutex is already held,
+      // and LK mutex isn't re-entrant safe.
+      if (ev->node.prev != nullptr && ev->node.next != nullptr) {
+        list_delete(&ev->node);
+      }
```
silverneko, line 79: s//delete_if_in_list

lib/uefi/events.cpp
silverneko: oh you've handled this already, nvm

lib/uefi/events.cpp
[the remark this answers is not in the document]
zhangxp1998: I would like to add a bit of safety net in code. If for whatever reason(logical error, unexpected edge cases), some events have their callback called but still not removed from the list, we can safely remove these stale events from the pending list, preventing leaked resources.

lib/uefi/events.cpp
zhangxp1998: Done, added deletion logic

lib/uefi/events.cpp
```diff
@@ -30,30 +31,57 @@
...
   }
+  return false;
+}
 
-  template <typename Func>
-  void write_access(Func f) {
-    AutoLock a(&m);
-    f(arr, events);
+bool invoke_callback_if_ready(EfiEventImpl *ev) {
+  if (ev->ready()) {
```
silverneko, line 56: and we just zapped the last user of `invoke_callback_if_ready`...

If no future use, then remove?

lib/uefi/events.cpp
```diff
@@ -30,30 +31,57 @@
...
+
+bool invoke_callback(EfiEventImpl *ev) {
+  DEBUG_ASSERT(ev->ready());
+  DEBUG_ASSERT(ev->creator_thread == get_current_thread());
+  if (!ev->callback_called && ev->notify_fn != nullptr) {
+    LTRACEF("Triggering event %p callback %p ctx %p on thread %s\n", ev,
+            ev->notify_fn, ev->notify_ctx, get_current_thread()->name);
+    backoff_wait_time = 0;
+    ev->notify_fn(ev, ev->notify_ctx);
+    ev->callback_called = true;
```
silverneko, line 49: here is also a potential race.

If notify_fn blocks, then it could get context switch to another thread. The other thread might call notify_fn again (re-entrant) since `callback_called` is still false.

this scenario is not possible now as we only allow the callback be called from the creator thread. but we need to keep this in mind when expanding the events API in the future to support multiple concurrent threads.

A potential improvement can replace `callback_called` with an atomic compare-and-swap of the `notify_fn` pointer, and then invoke the pointer. This ensures at most one thread can acquire the callback function pointer.

lib/uefi/events.cpp
zhangxp1998: Done

lib/uefi/events.cpp
zhangxp1998: I moved `ev->callback_called = true;` before calling `ev->notify_fn` to mitigate such risk. Since we only call `invoke_callback` after a node is already removed from pending list, and deleting a node from pending list is protected by a mutex, it is guaranteed that only 1 thread can call the callback function.

I've also added a TODO for using compare_and_swap, just to be extra cautious. LK doesn't have an atomic compare_and_swap function right now, so support must be added later.

lib/uefi/memory_protocols.cpp
```diff
@@ -45,14 +48,35 @@ mspace create_mspace_with_base_limit(void *base, size_t capacity, int locked) {
...
+  set_boot_aspace();
+  get_mspace();
+}
+
+void reset_heap() {
+  restore_aspace();
+  destroy_mspace(space);
+  space = nullptr;
+  free_pages(heap, kHeapSize / PAGE_SIZE);
+  heap = nullptr;
```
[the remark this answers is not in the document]
zhangxp1998: You are right on `restore_aspace` being last one, as `free_pages` would need the aspace. re-ordered. We can free the aspace after calling `free_pages`. Added that call as well.

lib/uefi/uefi.cpp
```diff
@@ -122,19 +124,27 @@ int load_sections_and_execute(bdev_t *dev,
   table.con_out = &console_out;
   table.configuration_table =
       reinterpret_cast<EfiConfigurationTable *>(alloc_page(PAGE_SIZE));
+  DEFER { free_pages(table.configuration_table, 1); };
```
silverneko, line 127: ditto. this would be called after `reset_heap()`

lib/uefi/memory_protocols.cpp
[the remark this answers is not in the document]
silverneko: hmm I think still not quite right. IIRC destructor is called inverted order. So `reset_heap` would be called before the other `free_pages` calls. The set/reset heap functions should be move to the beginning of the scope, or up a scope-level, or move to the caller.

PR #457, opened 2025-08-04
[ci][github] Disable buffering on subprocess output

zhangxp1998:
Testing script works by reading output of subprocess and checking if output matches expectation. To ensure that we can see output in timely fashion, disable buffering.


travisg (approved):
LGTM

PR #459, opened 2025-08-06
[lib][uefi] Move several memory APIs to uefi_platform.cpp

zhangxp1998:
This allows OEMs to customize exactly how their memories are allocated. For examples, OEMs might want to allocate certain parts of the UEFI from high address kernel space, for ease of access in LK world. Since UEFI spec requires everything to be identity mapped, the default implementation of memory allocation is identity mapped.

### DIFF
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -30,7 +30,6 @@
 #include <uefi/protocols/loaded_image_protocol.h>
 #include <uefi/types.h>
 
-#include "arch/defines.h"
 #include "blockio2_protocols.h"
 #include "blockio_protocols.h"
 #include "events.h"
@@ -112,23 +111,6 @@ EfiStatus locate_protocol(const EfiGuid *protocol, void *registration,
   return UNSUPPORTED;
 }
 
-EfiStatus allocate_pages(EfiAllocatorType type, EfiMemoryType memory_type,
-                         size_t pages, EfiPhysicalAddr *memory) {
-  if (memory == nullptr) {
-    return INVALID_PARAMETER;
-  }
-  if (type == ALLOCATE_MAX_ADDRESS && *memory < 0xFFFFFFFF) {
-    printf("allocate_pages(%d, %d, %zu, 0x%llx) unsupported\n", type,
-           memory_type, pages, *memory);
-    return UNSUPPORTED;
-  }
-  *memory = reinterpret_cast<EfiPhysicalAddr>(alloc_page(pages * PAGE_SIZE));
-  if (*memory == 0) {
-    return OUT_OF_RESOURCES;
-  }
-  return SUCCESS;
-}
-
 EfiStatus uninstall_multiple_protocol_interfaces(EfiHandle handle, ...) {
   printf("%s is unsupported\n", __FUNCTION__);
   return UNSUPPORTED;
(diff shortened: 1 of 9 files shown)


lib/uefi/boot_service_provider.cpp
silverneko, line 1: Could you also move `allocate_pages()` to memory_protocol.cpp. I think it makes sense to belong next to `allocate_pool`

lib/uefi/uefi_platform.cpp
```diff
@@ -109,3 +111,68 @@ __WEAK EfiStatus get_timestamp_properties(EfiTimestampProperties *properties) {
...
+  return "unknown";
+}
+template <typename T> T clamp(T n, T lower, T upper) {
+  if (n < lower) {
+    return lower;
+  }
+  if (n > upper) {
+    return upper;
+  }
+  return n;
```
silverneko, line 153: [no action required in this commit]

In a future cleanup pass, consider moving this to libcpp ?

lib/uefi/memory_protocols.cpp
silverneko, line 1: Also allow customizing `allocate_pool`, `uefi_malloc` and `free_pool` ?

lib/uefi/memory_protocols.h
```diff
@@ -10,10 +10,11 @@ vmm_aspace_t *set_boot_aspace();
...
 
-void *alloc_page(void *addr, size_t size, size_t align_log2 = PAGE_SIZE_SHIFT);
-void *alloc_page(size_t size, size_t align_log2 = PAGE_SIZE_SHIFT);
-EfiStatus free_pages(EfiPhysicalAddr memory, size_t pages);
-EfiStatus free_pages(void *memory, size_t pages);
+void *default_alloc_page(void *addr, size_t size,
+                         size_t align_log2 = PAGE_SIZE_SHIFT);
+void *default_alloc_page(size_t size, size_t align_log2 = PAGE_SIZE_SHIFT);
+EfiStatus default_free_pages(EfiPhysicalAddr memory, size_t pages);
+EfiStatus default_free_pages(void *memory, size_t pages);
```
silverneko, line 17: I think these shouldn't have external linkage. These are default implementations of `alloc_page` and `free_page`. Users should always call `alloc_page` and `free_page` directly, and `default_alloc_page` and `default_free_page` should never be called directly, thus should not be exported.

I suggest moving these function definitions to usefi_platform.cpp and inside a nameless namespace block so they have internal linkage.

zhangxp1998: Then we basically erased memory_protocols.cpp, as all of these functions are moved. I kept them here so that we can still organize code by their functionalities(memory/block/relocation)

lib/uefi/memory_protocols.cpp
zhangxp1998: For current use cases, customizing allocate_page is sufficient, as `allocate_pool` / `uefi_malloc` uses memory allocated by `allocate_pages` . So changing `allocate_pages` to allocate from kernel space would achieve the goal of making everything kernel space.

lib/uefi/memory_protocols.h
silverneko: I see. keeping `alloc_page` implementation in memory_protocol.cpp is fine, but `default_alloc_page` shouldn't have external linkage. Can you just provide a default (weak) implementation of `alloc_page` that just calls `default_alloc_page` (internal linkage / static / anonymous namespace), or just rename `default_alloc_page` to weak symbol definition of `alloc_page`:

uefi_platform.h:
```
// declare
void *alloc_page();
```

memory_protocol.h:
```
// move declare to uefi_platform.h
// void *alloc_page();
```

memory_protocols.cpp:
```
#include <uefi_platform.h>

WEAK void *alloc_page() {
  // default implementation
}
```

lib/uefi/memory_protocols.cpp
zhangxp1998: Just marked them as __WEAK anyway : )

lib/uefi/memory_protocols.cpp
```diff
@@ -230,4 +236,21 @@ EfiStatus get_physical_memory_map(size_t *memory_map_size,
   return SUCCESS;
 }
 
+EfiStatus allocate_pages(EfiAllocatorType type, EfiMemoryType memory_type,
+                         size_t pages, EfiPhysicalAddr *memory) {
+  if (memory == nullptr) {
+    return INVALID_PARAMETER;
+  }
+  if (type == ALLOCATE_MAX_ADDRESS && *memory < 0xFFFFFFFF) {
+    printf("allocate_pages(%d, %d, %zu, 0x%llx) unsupported\n", type,
```
silverneko, line 245: nit:

s/allocate_pages/__FUNCTION__



why not just use LTRACE macro?

lib/uefi/memory_protocols.cpp
```diff
@@ -230,4 +236,21 @@ EfiStatus get_physical_memory_map(size_t *memory_map_size,
   return SUCCESS;
 }
 
+EfiStatus allocate_pages(EfiAllocatorType type, EfiMemoryType memory_type,
```
silverneko, line 239: really minor nit: move next to `alloc_page` to group related functions together

lib/uefi/memory_protocols.h
```diff
@@ -10,14 +10,17 @@ vmm_aspace_t *set_boot_aspace();
...
+                         size_t pages, EfiPhysicalAddr *memory);
+
+// Declaration moved to uefi_platform.h to keep all weak functions declared in
+// one place.
+// void *uefi_malloc(size_t size);
+// EfiStatus allocate_pool(EfiMemoryType pool_type, size_t size, void **buf);
+// EfiStatus free_pool(void *mem);
+// EfiStatus free_pages(void *memory, size_t pages);
+
 void *identity_map(void *addr, size_t size);
```
silverneko, line 23: remove `identity_map` prototype declaration as the implementation is removed already

lib/uefi/memory_protocols.h
silverneko, line 1: nit: add license statement

lib/uefi/uefi_platform.cpp
```diff
@@ -25,6 +25,7 @@
 #include <uefi/types.h>
 
 #include "defer.h"
+#include "uefi/protocols/gbl_efi_image_loading_protocol.h"
```
silverneko, line 28: s/"/<

since we've setup the ipath, we should use `#include` with `<>`

silverneko (approved):
LGTM

just some minor nits

lib/uefi/memory_protocols.cpp
zhangxp1998: Done

lib/uefi/uefi_platform.cpp
zhangxp1998: Done

PR #460, opened 2025-08-08
[lib][uefi] Implement BootService methods

silverneko:


lib/uefi/uefi_platform.h
```diff
@@ -47,6 +47,9 @@ uint64_t get_timestamp();
 
 EfiStatus get_timestamp_properties(EfiTimestampProperties *properties);
 
+EfiStatus set_watchdog_timer(size_t timeout, uint64_t watchdog_code,
```
zhangxp1998, line 50: document unit of `timeout` ?

silverneko: done

lib/uefi/uefi_platform.cpp
```diff
@@ -111,6 +116,43 @@ __WEAK EfiStatus get_timestamp_properties(EfiTimestampProperties *properties) {
   return SUCCESS;
 }
 
+__BEGIN_CDECLS
+
+// If the target platform did not provide these methods, then the WEAK symbols
+// could resolve to NULL.
+extern __WEAK status_t platform_watchdog_init(
```
zhangxp1998, line 123: I just realized that LK already declares `platform_watchdog_init`

https://github.com/littlekernel/lk/blob/17dc5cd0aafdebd1da32f13fc6883c712a164f84/lib/watchdog/include/lib/watchdog.h#L60

What is the reason that we re-delcare them here with an additional "__WEAK" modifier?

silverneko: Updated code comment and commit message

PR #462, opened 2025-08-11
lib: uefi: implement volatile UEFI variables

grandpaul:
GBL needs to read an UEFI variable called "gbl_debug".
We add 2 commands. "uefi_set_var" to set an UEFI variable. And
"uefi_list_var" to list the UEFI variables.
    
For GBL, we can use "uefi_set_var gbl_debug 1" to enable its
debug mode.

### DIFF
--- a/lib/uefi/runtime_service_provider.cpp
+++ b/lib/uefi/runtime_service_provider.cpp
@@ -22,12 +22,19 @@
 #include <string.h>
 #include <uefi/types.h>
 
+#include "variable_mem.h"
+
 namespace {
 
 constexpr auto &&kSecureBoot = "SecureBoot";
 
 EFI_STATUS GetVariable(char16_t *VariableName, EfiGuid *VendorGuid,
                        uint32_t *Attributes, size_t *DataSize, void *Data) {
+
+  if (!VariableName || !VendorGuid || !DataSize) {
+    return INVALID_PARAMETER;
+  }
+
   char buffer[512];
   size_t i = 0;
   while (VariableName[i] && i < sizeof(buffer)) {
@@ -48,6 +55,21 @@ EFI_STATUS GetVariable(char16_t *VariableName, EfiGuid *VendorGuid,
     return SUCCESS;
   }
 
+  char *data_in_mem;
+  size_t data_in_mem_size;
+  if (efi_get_variable(VariableName, VendorGuid, Attributes, &data_in_mem, &data_in_mem_size) == SUCCESS) {
+    if (*DataSize == 0 && !Data) {
+      *DataSize = data_in_mem_size;
+      return BUFFER_TOO_SMALL;
+    }
+    if (data_in_mem_size > *DataSize) {
+      return BUFFER_TOO_SMALL;
+    }
+    *DataSize = data_in_mem_size;
+    memcpy(Data, data_in_mem, data_in_mem_size);
+    return SUCCESS;
+  }
+
   printf("%s(%s) is unsupported\n", __FUNCTION__, buffer);
   return UNSUPPORTED;
 }
@@ -61,7 +83,23 @@ EFI_STATUS SetVirtualAddressMap(size_t MemoryMapSize, size_t DescriptorSize,
 
 EFI_STATUS SetVariable(char16_t *VariableName, EfiGuid *VendorGuid,
                        uint32_t Attributes, size_t DataSize, void *Data) {
-  printf("%s is unsupported\n", __FUNCTION__);
+  if (!VariableName || VariableName[0] == 0) {
+    return INVALID_PARAMETER;
+  }
+
+  /* Only allow setting non-volatile variables */
+  if ((Attributes & EFI_VARIABLE_NON_VOLATILE) == 0) {
+    efi_set_variable(VariableName,
+                     VendorGuid,
+                     Attributes,
+                     reinterpret_cast<const char *>(Data),
+                     DataSize);
+    return SUCCESS;
+  }
+
+  printf("%s: Only non-volatile is supported. Attributes = 0x%x\n",
+         __FUNCTION__,
+         Attributes);
   return UNSUPPORTED;
 }
 
--- a/lib/uefi/variable_mem.cpp
+++ b/lib/uefi/variable_mem.cpp
@@ -1,0 +1,176 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


lib/uefi/variable_mem.h
```diff
@@ -0,0 +1,46 @@
...
+            0x93ca,
+            0x11d2,
+            {0xaa, 0x0d, 0x00, 0xe0, 0x98, 0x03, 0x2b, 0x8c}};
+
+static constexpr uint32_t EFI_VARIABLE_NON_VOLATILE = 0x01;
+static constexpr uint32_t EFI_VARIABLE_BOOTSERVICE_ACCESS = 0x02;
+static constexpr uint32_t EFI_VARIABLE_RUNTIME_ACCESS = 0x04;
+static constexpr uint32_t EFI_VARIABLE_HARDWARE_ERROR_RECORD = 0x08;
+
+EfiStatus efi_get_variable_mem(const char *variable_name,
```
zhangxp1998, line 34: Just use `efi_get_variable_mem` , we don't need to emphasize that this is implemented in memory

lib/uefi/runtime_service_provider.cpp
```diff
@@ -61,6 +83,25 @@ EFI_STATUS SetVirtualAddressMap(size_t MemoryMapSize, size_t DescriptorSize,
 
 EFI_STATUS SetVariable(char16_t *VariableName, EfiGuid *VendorGuid,
                        uint32_t Attributes, size_t DataSize, void *Data) {
+  if (!VariableName || VariableName[0] == 0) {
+    return INVALID_PARAMETER;
+  }
+
+  /* Only allow setting non-volatile variables */
+  if ((Attributes & (~(EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_RUNTIME_ACCESS))) == 0) {
```
zhangxp1998, line 91: Check if volatile flag is set, then print a message saying only non-volatile variable is supported, and return UNSUPPORTED in volatile branch.

lib/uefi/runtime_service_provider.cpp
```diff
@@ -61,6 +83,25 @@ EFI_STATUS SetVirtualAddressMap(size_t MemoryMapSize, size_t DescriptorSize,
...
+    char buffer[512];
+    memset(buffer, 0, sizeof(buffer));
+    for (size_t i = 0; i < sizeof(buffer) && VariableName[i] != 0; i++) {
+      buffer[i] = VariableName[i];
```
zhangxp1998, line 95: this has happened enough times that I think we should add a `utf16_to_utf8` helper function, even if that helper function currently only handles simple ascii chars.

lib/uefi/variable_mem.cpp
```diff
@@ -0,0 +1,198 @@
...
+    EfiGuid VendorGuid;
+    uint32_t Attributes;
+    char* Data;
+    size_t DataLen;
+  };
+}
+
+static list_node variables_in_mem = LIST_INITIAL_VALUE(variables_in_mem);
+
+static int comp_variable_name(const char16_t *varname1,
```
zhangxp1998, line 39: we should be able to use memcmp

lib/uefi/variable_mem.cpp
```diff
@@ -0,0 +1,198 @@
...
+  /* alloc new variable if it is not existed */
+  if (!var) {
+    var = reinterpret_cast<struct EfiVariableInMem *>(malloc(sizeof(struct EfiVariableInMem)));
+    memset(var, 0, sizeof(struct EfiVariableInMem));
+    var->node = LIST_INITIAL_CLEARED_VALUE;
+    int name_len = (strlen(variable_name) + 1) * 2;
+    var->VariableName = reinterpret_cast<char16_t *>(malloc(name_len));
+    memset(var->VariableName, 0, name_len);
+    for (size_t i = 0; i < strlen(variable_name) ; i++) {
+      var->VariableName[i] = variable_name[i];
```
zhangxp1998, line 116: caller just converted utf16 to utf8, and now we are converting back to utf16, I don't quite understand why we are doing this. Maybe just change `efi_set_variable_mem` to take utf16 directly?

zhangxp1998 (changes requested):

lib/uefi/variable_mem.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include <malloc.h>
+#include <lk/list.h>
+#include <uefi/types.h>
+#include "charset.h"
+
+namespace {
+  struct EfiVariableInMem {
```
zhangxp1998, line 28: Remove the `InMem` suffix, as structs are inherently in memory, and we don't have persistent variables anyway.

lib/uefi/runtime_service_provider.cpp
```diff
@@ -61,7 +83,21 @@ EFI_STATUS SetVirtualAddressMap(size_t MemoryMapSize, size_t DescriptorSize,
...
+  if ((Attributes & EFI_VARIABLE_NON_VOLATILE) == 0) {
+    efi_set_variable(VariableName,
+                     VendorGuid,
+                     Attributes,
+                     reinterpret_cast<const char *>(Data),
+                     DataSize);
+    return SUCCESS;
+  }
+
+  printf("%s: Only non-volatile is supported\n", __FUNCTION__);
```
zhangxp1998, line 100: print value of `Attributes` for ease of debugging.

lib/uefi/variable_mem.cpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+void efi_set_variable(const char16_t *variable_name,
+                      const EfiGuid *guid,
+                      uint32_t attribute,
+                      const char *data,
+                      size_t data_len) {
+  struct EfiVariable *var = nullptr;
+
+  var = search_existing_variable(variable_name, guid);
```
zhangxp1998, line 92: declare and define in 1 step, merge L90 and L92

lib/uefi/runtime_service_provider.cpp
```diff
@@ -22,12 +22,19 @@
...
 constexpr auto &&kSecureBoot = "SecureBoot";
 
 EFI_STATUS GetVariable(char16_t *VariableName, EfiGuid *VendorGuid,
                        uint32_t *Attributes, size_t *DataSize, void *Data) {
+
+  if (!VariableName || !VendorGuid || !DataSize) {
+    return INVALID_PARAMETER;
+  }
+
   char buffer[512];
```
zhangxp1998, line 38: You can do this in a later CL

replace this with the newly added charset functions

PR #465, opened 2025-08-20
[arch][x86] Read full 32bit apic id from x2apic msr if available

2544632b:


arch/x86/mp.c
```diff
@@ -87,18 +88,29 @@ status_t arch_mp_send_ipi(mp_cpu_mask_t target, mp_ipi_t ipi) {
...
-
-    // TODO: read full 32bit apic id from x2apic msr if available
-
-    return apic_id;
+    uint32_t unused = 0, ecx;
+    cpuid(0x1, &unused, &unused, &ecx, &unused);
+
+    if (ecx & (1u << 21)) {
+        // read full 32bit apic id from x2apic msr if available
+        return (uint32_t)read_msr(0x802);
```
travisg, line 96: I like the idea of this. I think it might be a bit cleaner if it made a call into the lapic.c code which also is responsible for bringing up the x2apic per cpu to request the apic id there. This way it can initialize the x2apic without needing to do it again here.

I think basically moving this code into the lapic.c file would suffice, and have it automatically enable x2apic mode if it's called would work.

travisg (changes requested):
Looks good, just some organizational changes requested.

Thanks!

2544632b:
🤔

travisg:
Hmm, that seems to be better. Need to run some test cases on the x86 hardware I have around here to make sure everything is good but should be able to take it.

PR #468, opened 2025-08-26
[lib][uefi] Fix memory leak in load_pe_file

zhangxp1998:


travisg:
Oh huh, I didn't know about this Copilot review thing, it must have been kicked in. Well, take what you need out of it, there's probably some actual useful stuff in there.

PR #469, opened 2025-09-04
[lib][uefi] Update UEFI protocol definitions

silverneko:
GBL HEAD: https://android.googlesource.com/platform/bootable/libbootloader/+/878653395db0a046f98a7d6c6efb46afd45738b8


zhangxp1998 (approved):

travisg:
I should probably set the CI builds to use -Werror, but i can't set it by default because I dont control the toolchains that folks use. However, this change adds a warning with GCC concerning uchar_t:

`
compiling lib/uefi/uefi.cpp
In file included from lib/uefi/include/uefi/protocols/device_path_protocol.h:29,
                 from lib/uefi/include/uefi/boot_service.h:29,
                 from lib/uefi/uefi.cpp:30:
lib/uefi/include/uefi/protocols/types.h:37:17: warning: identifier ‘char8_t’ is a keyword in C++20 [-Wc++20-compat]
   37 | typedef uint8_t char8_t;
`

travisg:
If you want to locally compile with -Werror all the time, add
`WERROR:=1` to local.mk and put it in your root, or just set WERROR environment variable.

silverneko:
Fix warning with #471

PR #472, opened 2025-09-05
[lib][uefi] Implement EFI_ERASE_BLOCK protocol

silverneko:
Supports only synchronous IO for now.


silverneko:
@zhangxp1998 PTAL

PR #473, opened 2025-09-16
[lib][uefi] Update protocol headers from upstream

zhangxp1998:
Copied from https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/libefi_types/defs/protocols/


lib/uefi/include/uefi/protocols/timestamp.h
silverneko, line 1: Could you also remove the duplicated file lk/lib/uefi/include/uefi/protocols
/efi_timestamp_protocol.h

lib/uefi/include/uefi/protocols/block_io_protocol.h
```diff
@@ -13,25 +13,36 @@
...
 #ifndef __BLOCK_IO_PROTOCOL_H__
 #define __BLOCK_IO_PROTOCOL_H__
 
+#include "gbl_protocol_utils.h"
+#include "types.h"
+
 typedef struct EfiBlockIoMedia EfiBlockIoMedia;
 typedef struct EfiBlockIoProtocol EfiBlockIoProtocol;
 
+static constexpr uint64_t EFI_BLOCK_IO_PROTOCOL_REVISION =
```
silverneko, line 35: What's the reason to replace `const` with `constexpr`? 

Can we keep it as the C-compatible `const` as some vendors might be writing pure C?

lib/uefi/include/uefi/protocols/timestamp.h
zhangxp1998: Done

lib/uefi/include/uefi/protocols/block_io_protocol.h
zhangxp1998: Done

lib/uefi/include/uefi/types.h
```diff
@@ -31,15 +31,13 @@
 
 #include "gbl_protocol_utils.h"
 
+#ifndef __cplusplus
+#define const const
+#endif
```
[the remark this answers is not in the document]
zhangxp1998: removed

PR #474, opened 2025-09-18
Migrate to latest GBL protocols

zhangxp1998:


lib/uefi/memory_protocols.cpp
```diff
@@ -27,8 +26,12 @@
 #include <string.h>
 #include <sys/types.h>
 #include <uefi/boot_service.h>
+#include <uefi/protocols/gbl_efi_boot_memory_protocol.h>
 #include <uefi/types.h>
 
+#include "uefi/protocols/types.h"
```
silverneko, line 32: this line is redundant as it is the same as `<uefi/types.h>`

silverneko (approved):

lib/uefi/memory_protocols.cpp
zhangxp1998: Done

silverneko (approved):

PR #478, opened 2025-10-13
[lib][uefi] implement UEFI HII Database Protocol

grandpaul:
Shell.efi from EDK2 requires this protocol to be existed. Otherwise it crashes. So we implemented a minimum set of UEFI HII Database Protocol to let the Shell.efi to run.

### DIFF
--- a/lib/uefi/hii_protocol.cpp
+++ b/lib/uefi/hii_protocol.cpp
@@ -1,0 +1,268 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


travisg:
I'm okay with this but @zhangxp1998 has generally been the maintainer of the UEFI stuff. What do you think Kelvin?

lib/uefi/hii_protocol.cpp
```diff
@@ -0,0 +1,268 @@
...
+    for (uint32_t i = 0; i < stbl->nstrings; i++)
+      free(stbl->strings[i].string);
+    free(stbl->strings);
+    free(stbl->language);
+    free(stbl);
+  }
+
+  static void remove_strings_package(struct efi_hii_packagelist *hii) {
+    while (true) {
+      struct efi_string_table *stbl;
```
zhangxp1998: Define and initialize in 1 step.

lib/uefi/hii_protocol.cpp
```diff
@@ -0,0 +1,268 @@
...
+ */
+
+#include <stdio.h>
+#include <stdint.h>
+#include <string.h>
+#include <malloc.h>
+#include <uefi/protocols/hii_protocol.h>
+
+namespace {
+  const int EFI_HII_PACKAGE_TYPE_SHIFT = 24;
```
zhangxp1998: constexpr

zhangxp1998:
Superseded by #524, which rebases this work onto current master and addresses the review feedback.

PR #479, opened 2025-10-14
The beginnings of Rust support for LK

d3zd3z:
This adds the basics of Rust support for lk. Notably, it includes:

- A pl011 driver written in Rust.
- Interfaces to init, log, cbuf in lk.
- A type 'LkOnce' that provides Once-like functionality but in a way that works with LK's init mechanism.
- Rust code is built as a single build with `cargo` to generate a .a file linked into the build.

The Rust support is not enabled by default, and can be enabled by passing `USE_RUST=1` to the make invocation. Currently only qemu-virst-arm64-test is supported, and this has only been tested when using clang to compile the C code.


travisg:
Starting a review now, uno momento.

rust/dev-pl011/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition = "2024"
+description = """
+Rust PL011 uart driver for LK.
+"""
+
+[dependencies.lk]
+version = "0.1.0"
+path = "../lk"
+
+[dependencies.log]
```
travisg: Question: what are the licenses of these two libs. Are they compatible with LKs?

rust/dev-pl011/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    fn get_dev(&self) -> UniqueMmioPointer<'static, UartRegs> {
+        unsafe { UniqueMmioPointer::new(NonNull::new(self.config.base as _).unwrap()) }
+    }
+}
+
+const RXBUF_SIZE: usize = 32;
+
+static PL011_UARTS: [LkOnce<Pl011Uart>; 1] = [LkOnce::uninit()];
+
+extern "C" fn uart_irq(arg: *mut c_void) -> sys::handler_return {
```
travisg: super cool that this just directly lines up with the irq model.

rust/dev-pl011/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+#[unsafe(no_mangle)]
+extern "C" fn uart_getc(port: c_int, wait: bool) -> c_int {
+    let uart = get_uart(port as usize);
+    let mut dev = uart.get_dev();
+    let buf = unsafe { Cbuf::new(uart.rx_buf.get()) };
+
+    if let Some(ch) = buf.read_char(wait) {
+        // TODO: This modify is racy, as the interrupt handler also does a
```
travisg: Yah this will need to be wrapped in a spinlock here and in the IRQ to make that go away, I think we discussed something like that the other day.

It's okay for now though, this driver can go through some more clean up to work out a lot of details.

rust/lk-sys/build.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    // Allow `SYS_INCLUDES` to be specified to specify a path where the system includes for this
+    // version of clang.
+    let bindings = if let Ok(sys_inc) = env::var("SYS_INCLUDE") {
+        bindings.clang_args(&["-isystem", &sys_inc])
+    } else {
+        bindings
+    };
+
+    let bindings = bindings
```
travisg: Is this going to be the basic mechanism for all things imported from LK? They'll have to be called out here?

If so, some future thought may be to have each module declare some sort of rust export list and consolodate it here maybe?

rust/lk-sys/build.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    Ok(())
+}
+
+/// Convert the provided include path to properly work.  There are two issues here:
+/// The GLOBAL_INCLUDES variable contains paths with a leading `-I` on them, which we need to strip.
+/// In addition, the paths are relative to the top of the build tree.  Cargo invokes the build.rs
+/// code while within the crate directory.
+///
+/// For now, we'll assume we are two directories deep and build the paths appropriately.
```
travisg: Yah as pointed out before this will be an issue eventually, with the split LK repo builds, though I think for now we can go with this, but maybe see if you can make it hard fail if it sees a path in global includes that doesn't work? Dunno if that's easily possible.

rust/lk-sys/rules.mk
```diff
@@ -0,0 +1,43 @@
...
+MODULE := $(LOCAL_DIR)
+
+# Rust specific module support.
+# Although this starts here, this should be generalized, possibly into a module_rust.mk.
+
+MODULE_SRCDIR := $(MODULE)
+MODULE_BUILDDIR := $(call TOBUILDDIR,$(MODULE_SRCDIR))
+
+# TODO These should be set by the arch.
+RUST_TARGET := aarch64-unknown-none
```
travisg: TODO: figure out how we can define our own triple and build the core/alloc crates ourselves, as we talked about a bit the other day. The generation of floating point code with the default triple is already technically a major problem as it stands, since there's nothing that keeps LLVM from generating floating point code in the pl011 irq handler with the standard triple.

rust/lk/src/log.rs
```diff
@@ -0,0 +1,105 @@
...
+        4 => LevelFilter::Debug,
+        _ => LevelFilter::Trace, // enable trace! at 5+
+    });
+}
+
+// LK_INIT_HOOK!(KERNEL_LOG_INIT, kernel_log_init_func, lk_init_level::LK_INIT_LEVEL_HEAP);
+LK_INIT_HOOK!(
+    KERNEL_LOG_INIT,
+    kernel_log_init_func,
+    lk_init_level::LK_INIT_LEVEL_ARCH
```
travisg: I'm guessing this could actually be somewhat earlier, probably INIT_LEVEL_THREADING?

rust/lk/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+pub mod init;
+pub mod lkonce;
+pub mod log;
+pub mod macros;
+
+/// Initialize the rust side of the code.
+///
+/// Due to linker optimization, without a direct call into this code, the linker will not bring in
+/// anything.
+pub fn init() {
```
travisg: Though it's not doing much, I think i mentioned it in the platform code, we should probably call this as soon as possible, so that any rust code that comes online later can just assume any background rust setup has already happened. OTOH, this fputs may be assuming there's already some sort of log path available? May need to init silently.

platform/qemu-virt-arm/platform.c
```diff
@@ -87,9 +87,17 @@ void platform_early_init(void) {
...
+#if USE_RUST
+void setup_rust(void);
+#else
+static void setup_rust(void) {}
+#endif
+
 void platform_init(void) {
     pl011_init(0);
 
+    setup_rust();
```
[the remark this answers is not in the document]
travisg: Also in general it'd be good to have the platform code not direclty involved in any setting up of rust, except maybe to manually initialize the pl011 where it's needed, as it does here.

rust/dev-pl011/src/lib.rs
d3zd3z: I do have a desire, though, to create a proc macro to make this a little easier to declare, more along the lines of something like:
```rust
#[lk::irq_handler]
fn uart_irq(arg: *mut c_void) -> sys::handler_return ...
```
but I'm not actually sure if that is all that helpful. More helpful would be something a bit more strongly typed so that the type of the handler argument would be a real rust type instead of having to see the `*mut c_void` from the rust side.

That kind of comes down to the question of how much we want to abstract the lk C mechanisms from the Rust code. And that doesn't have to be decided now, as we do more, it might become more clear what is actually wanted.

rust/lk-sys/build.rs
d3zd3z: That's a good idea, and I'll think about the best mechanism for that. That list might grow a bit beyond what makes sense for an environment variable, but the rust support rules.mk could put the result into a file that build.rs reads.

I just started with a template from what trusty did.

It isn't quite clear to me where this list should be, though. Perhaps it should be declared in the Rust modules that wrap the LK functionality, although those don't currently have any makefiles that get read. But, that should probably change, and instead of `rust_support` directly referencing the various crates, we could have each crate have a rules.mk, which registers the crate. Then `rust_support` would build a list of the dependencies up when generating it's files instead of them having to be hard coded. That would make the choice of modules determined more like how it is done within the rest of lk.

rust/lk/src/macros.rs
```diff
@@ -0,0 +1,77 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/// Helper macro for container_of and container_of_mut. Exported so it can be
```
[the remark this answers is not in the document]
d3zd3z: Although, amusingly, these particular comments just came from trustee.

However, I would like to get clippy involved when we build this in CI. I can start by a declaration at the top that changes missing doc comments into a compile error. It doesn't care if the comments are helpful or meaningful, but at least flags missing ones.

rust/lk/src/lib.rs
d3zd3z: Realistically, this probably shouldn't even be called 'init', since it isn't actually doing anything, just making sure the code is linked in. The init can happen through the normal lk init mechanism.

.gitignore
```diff
@@ -19,3 +19,5 @@ tags
 toolchain
 .cache
 vmlinux
+target/
```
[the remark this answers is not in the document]
d3zd3z: That's a mistake. I wanted to ignore target directories in some of the subdirectories of rust crates because sometimes rust analyzer will erroneously put one there, but we definitely don't want to ignore this at the top level.

lib/rust_support/rules.mk
```diff
@@ -0,0 +1,85 @@
...
+$(MODULE_OBJECT): MODULE_BUILDDIR:=$(MODULE_BUILDDIR)
+
+# Override with module local values for the build rule.
+$(MODULE_OBJECT): $(MODULE_OBJECT).phony $(MODULE_OBJS) $(CARGO_CONFIG)
+	cd $(MODULE_BUILDDIR); \
+		cargo build
+
+EXTRA_OBJS := $(EXTRA_OBJS) $(MODULE_OBJECT)
+
+# Bring in the source files via copy.
```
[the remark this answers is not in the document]
d3zd3z: I did end up having to dig around a bit to figure out if -p was supported in all of the places we'd want.

However, TESTANDREPLACE is still needed when the stub is generated, since the temp file will always be new.

rust/dev-pl011/Cargo.toml
d3zd3z: Look more deeply, the LICENSE file at the top _is_ actually the MIT text. So, I've put MIT headers on each of the files I've created (the ones from trusty seem to already have that), and added `license = "MIT"` lines to the Cargo.toml files.

rust/lk-sys/rules.mk
d3zd3z: I would also like to address doing this in a more complete way with an additional PR(s).

lib/rust_support/rules.mk
```diff
@@ -0,0 +1,85 @@
...
+
+$(CARGO_CONFIG).phony:
+.PHONY: $(CARGO_CONFIG).phony
+
+$(CARGO_CONFIG): MODULE_SRCDIR:=$(MODULE_SRCDIR)
+
+$(CARGO_CONFIG): $(CARGO_CONFIG).phony
+	mkdir -p $(dir $@)
+	@tmp="$@.tmp"; { \
+		echo "# AUTOGENERATED -- edit $(MODULE_SRCDIR)/rules.mk instead"; \
```
zhangxp1998: should this be 
```
		echo "# AUTOGENERATED -- edit $(MODULE_SRCDIR)/Cargo.toml.in instead?"; \
```

rust/lk/src/log.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::init::lk_init_level;
+
+use crate::lk_stderr;
+use crate::sys::fflush;
+use crate::sys::fputs;
+// use crate::sys::LK_LOGLEVEL_RUST;
+// TODO: Configure this properly.
+static LK_LOGLEVEL_RUST: usize = 4;
+
+static TRUSTY_LOGGER: TrustyKernelLogger = TrustyKernelLogger;
```
zhangxp1998: I wonder if we should avoid mentioning Trusty here, as LK is really its own project.

toor1245:
I have a question, does this mean that new functionality should be implemented using Rust first?

I'm asking because I would like to add MMC/SD card API support and first driver is pl180 for Arm FVP emulator, and there's an implementation written in C (WIP).
https://github.com/toor1245/lk/tree/sdcard

So I'm wondering if it's worth trying to rewrite it using Rust.

zhangxp1998:
> I have a question, does this mean that new functionality should be implemented using Rust first?
> 
> I'm asking because I would like to add MMC/SD card API support and first driver is pl180 for Arm FVP emulator, and there's an implementation written in C (WIP). https://github.com/toor1245/lk/tree/sdcard
> 
> So I'm wondering if it's worth trying to rewrite it using Rust.

This is not an attempt to replace C code in LK with rust, or enforce new code to be written in rust. This effort is only meant to enable Rust code for those who need.

travisg:
Looking at it now, so far everything is looking great, though I think I can tweak some of the build system stuff to make it a little more conditional and a few less cross-module deps in the C world. Would you mind if i just pushed a change or two on top of this stack?

PR #481, opened 2025-10-21
[platform][pc] multiboot2 support

2544632b:
not completely, I have not implemented the ACPI and framebuffer display.

### DIFF
--- a/lib/libc/string/rules.mk
+++ b/lib/libc/string/rules.mk
@@ -36,12 +36,7 @@ LIBC_STRING_C_DIR := $(LOCAL_DIR)
 # include the arch specific string routines
 #
 # the makefile may filter out implemented versions from the C_STRING_OPS variable
-# x86 or x86-64 or anothers .....? T_T
-ifeq ($(SUBARCH),)
-	-include $(LOCAL_DIR)/arch/$(SUBARCH)/rules.mk
-else
-	-include $(LOCAL_DIR)/arch/$(ARCH)/rules.mk
-endif
+-include $(LOCAL_DIR)/arch/$(ARCH)/rules.mk
 
 MODULE_SRCS += \
 	$(addprefix $(LIBC_STRING_C_DIR)/,$(addsuffix .c,$(C_STRING_OPS)))


arch/x86/arch.c
```diff
@@ -55,6 +55,9 @@ uint8_t _kstack[PAGE_SIZE] __ALIGNED(sizeof(unsigned long));
 /* save a pointer to the multiboot information coming in from whoever called us */
 /* make sure it lives in .data to avoid it being wiped out by bss clearing */
 __SECTION(".data") uint32_t _multiboot_info;
+/* Try get the system table from tag 12, if present */
+/* `#define MULTIBOOT2_TAG_TYPE_EFI64    12` */
+__SECTION(".data") uint64_t _efi64_system_table;
```
travisg: There shouldn't be a need to declare this in a special section since it's going to be overwritten by plain C code later.  The reason the _multiboot_info variable is declare this way is it's set in extremely early start.S code before it zeros out the BSS, so it has to be set aside so the BSS clearing doesn't wipe it out.

dev/include/hw/multiboot.h
```diff
@@ -172,5 +223,82 @@ enum {
...
 
+#if MULTIBOOT2_SUPPORT
+
+struct multiboot2_info {
+    uint32_t total_size;
+    uint32_t reserved;
+};
+
+struct multiboot2_tag {
+    uint32_t type;
```
travisg: The spec seems to indicate this is actually two 16bit type and flags fields. Since it's little endian, it just so happens that type is in the bottom 16 bits and the current code doesn't deal with any flags.

Either way may make sense to follow the spec and have two fields here.

dev/include/hw/multiboot.h
```diff
@@ -172,5 +223,82 @@ enum {
...
+    uint8_t framebuffer_bpp;
+    uint8_t framebuffer_type;
+    uint16_t reserved;
+};
+
+struct multiboot2_tag_framebuffer {
+    struct multiboot2_tag_framebuffer_common common;
+    union {
+        struct {
+            uint16_t framebuffer_palette_num_colors;
```
travisg: spec says this is a 32bit.

dev/include/hw/multiboot.h
```diff
@@ -172,5 +223,82 @@ enum {
...
+        };
```
travisg: suggestion: can we name these enums something to disambiguate them at run time? ie, fb.rgb.* or fb.palette.*?

lib/acpi_lite/acpi_lite.cpp
```diff
@@ -284,6 +342,8 @@ static status_t initialize_table(size_t i) {
 status_t acpi_lite_init(paddr_t rsdp_pa) {
   LTRACEF("passed in rsdp %#" PRIxPTR "\n", rsdp_pa);
 
+#if MULTIBOOT2_SUPPORT
```
travisg: nit: if !MULTIBOOT2_SUPPORT

platform/pc/platform.c
```diff
@@ -44,6 +44,9 @@ static bool found_acpi = false;
 
 /* multiboot information passed in, if present */
 extern uint32_t _multiboot_info;
+#if MULTIBOOT2_SUPPORT
+extern uint64_t _efi64_system_table;
```
travisg: we should probably move these two a header, perhaps in hw/multiboot.h here and in the acpi_lite code.

dev/include/hw/multiboot.h
2544632b: but GRUB used 16bit at here.
```
struct multiboot_tag_framebuffer
{
  struct multiboot_tag_framebuffer_common common;

  union
  {
    struct
    {
      multiboot_uint16_t framebuffer_palette_num_colors;
      struct multiboot_color framebuffer_palette[0];
    };
```

travisg: huh okay i guess if that's what grub does <shrug>. Maybe leave a comment to that effect.

travisg:
Drop a note when you think it's ready for another look.

2544632b:
hey @travisg , Is this good?

travisg:
Oh sorry, missed the review request. Will take a look in the next couple days. Double check that it compiles with and without the #defines set. Ultimately i'd like to work towards having it always on and seeing if it can support both natively but I think that can be some additional CLs after this one.

travisg:
If you're going to fiddle with the string routines can you do those on another PR? I'll get to this one pretty soon but probably dont want to mess it up with extra changes.

2544632b:
And about the multiboot2, multiboot1 will lose something when with ovmf so I implemented it.

travisg:
I did some research and I think the reason I evetually want to support both at the same time (which is totally feasible) is qemu -kernel still only supports multiboot 1, even if modern grub supports 2. But it seems easy enough to handle both, just some work in start.S to deal with multiple entry points (probably).

travisg:
Going to merge this in a moment.

travisg:
Merged everything up to the string function addition and removal. Thanks!

Will do a bit of work to clean up the display stuff, since there's already a lib/gfx console that can use, should be able to refactor this.

PR #483, opened 2025-10-29
[lib][uefi] Update protocol headers from upstream

silverneko:
https://android.googlesource.com/platform/bootable/libbootloader/+/16d3f61423eb100a43a327874da26e2012ce1559

### DIFF
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -54,16 +54,16 @@ EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
   if (guid_eq(protocol, LOADED_IMAGE_PROTOCOL_GUID)) {
     printf("handle_protocol(%p, LOADED_IMAGE_PROTOCOL_GUID, %p);\n", handle,
            intf);
-    const auto loaded_image = static_cast<EFI_LOADED_IMAGE_PROTOCOL *>(
-        uefi_malloc(sizeof(EFI_LOADED_IMAGE_PROTOCOL)));
+    const auto loaded_image = static_cast<EfiLoadedImageProtocol*>(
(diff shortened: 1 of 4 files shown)


silverneko:
@zhangxp1998 PTAL

lib/uefi/rules.mk
```diff
@@ -7,6 +7,9 @@ MODULE_DEFINES=MSPACES=1
 MODULE_INCLUDES += \
 	lib/watchdog/include \
 
+GLOBAL_INCLUDES += \
+	$(LOCAL_DIR)/include/uefi \
```
zhangxp1998: I don't think adding this dir to GLOBAL_INCLUDE is a good idea, it would pollute the global include namespace for everyone. For me it's better to take the work to patch include path each time, or create some symlinks so that files in `lib/uefi/include` can use #include "" properly. What do you think?

silverneko: Didn't `GLOBAL_INCLUDES ` take affect only if the module is in the `MODULE_DEPS` of the top level project? Does LK even have a "globel namespace"? I thought if a project does not depend on `lib/uefi` (directly or indirectly) then the global include path would not be modified at all.

> Within a module you can also add MODULE_INCLUDES that just works for it.

I've tried this but it didn't work. For example a new project "project/foo" uses `lib/uefi` and writes `#include <uefi/protocols/gbl_efi_fastboot_protocol.h‎>`. This would cause a build error bacause `project/foo` does not know about the `include/uefi` include path, thus the `#include "types.h"` clause in `gbl_efi_fastboot_protocol.h` would fail to resolve to `uefi/types.h`

silverneko: second thought, let me try to fix this from upstream instead.

zhangxp1998 (approved):

PR #484, opened 2025-11-02
[app][tests] Change 'buf_in' to unsigned char for cbuf_tests

jrac:
] cbuf_tests
running basic tests...
running random tests...
panic (caller 0x28e0): 128 != 4294967168 (app/tests/cbuf_tests.c:109)
CRASH: starting debug shell, reason 'software panic'
entering panic shell loop
! 

cbuf_tests fails on m68k due to 'char' being signed. This commit switches 'buf_in' to unsigned char.

Not sure if 'buf' and 'buf_out' should also be switched to unsigned; the test passes with changes to just buf_in.

### DIFF
--- a/app/tests/cbuf_tests.c
+++ b/app/tests/cbuf_tests.c
@@ -99,7 +99,7 @@ int cbuf_tests(int argc, const console_cmd_args *argv) {
 
         // Read up to 8 bytes, make sure they are right.
         if (pos_in < pos_out) {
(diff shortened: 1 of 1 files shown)


travisg:
Thanks, yeah this test seems to have just been broken for a while on most arches which have signed chars. I really should convert the test to the unittest framework so it gets run all the time, not just when manually invoked.

PR #492, opened 2025-12-27
[m68k] Integrate Motorola Integer Software Package (68060SP), for '060

jrac:
- Add 68060SP ISP code to external/arch/m68k
   - Install _060_isp_unimp handler for "Unimplemented Integer Instruction" exception
   - Add 68060 CPU target
   
Some design choices that need scrutiny:
   
   * The '_060_isp_unimp' routine is a weak symbol to allow things to build when the ISP code is not being assembled/linked.
   
   * In arch/m68k/rules.mk there is a conditional that performs a pattern match, rather than shelling out to numeric comparison or similar. The ISP code fails to build '010 targets like rosco, hence the gating.


travisg:
Looks good to me, i'll add a switch to see if can test this via qemu.

PR #496, opened 2026-02-28
[lib][gfx] Add mono 1bpp support

jrac:
[lib][font] Use mono 1bpp byte-span drawing, where available

- Adds 1bpp rendering function, and accompanying implementations of putpixel, copyrect, fillrect.

- libgfx now calls arch_clean_cache_range() conditionally, to avoid crashes on architectures lacking cache support.

- Includes changes made by scripts/codestyle


travisg (approved):
LGTM! Thanks!

PR #499, opened 2026-03-23
[arch][m68k] Add support for plain 68000 exception handling

jrac:
Generates per-vector stubs, with each vectory entry pointing to a unique stub slot, which branches to a common handler.

We construct a synthetic exception frame that can be passed to the existing C handlers.

### DIFF
--- a/arch/m68k/exceptions_asm.S
+++ b/arch/m68k/exceptions_asm.S
@@ -9,9 +9,6 @@
 
 .text
 
-.weak _060_isp_unimp
-_060_isp_unimp:
-
 #if M68K_CPU >= 68010
 // For 68010s and above the exception frame already has a code pushed on the stack,
 // so it's easy to just vector most of the handlers into one of a few classes of handlers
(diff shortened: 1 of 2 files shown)


arch/m68k/exceptions_asm.S
```diff
@@ -93,5 +93,162 @@ DATA(exc_vectors)
...
+    .rept (15 - 2)
+    .long stub
+    .set stub, stub + STUB_SIZE
+    .endr
+.org (15 * 4)
+    .set stub, _m68k_irq_stub_gen
+    .long stub
+    .set stub, stub + STUB_SIZE
+.org (16 * 4)
+    .set stub, _m68k_general_stub_gen
```
travisg: is this resetting the stub its referring to back to the first entry of the general stub table? i think this makes entry 16 point to entry 2, etc

travisg (commented):
I think there are a few issues here but otherwise LGTM.

travisg:
Sorry for taking a while, but this looks goot to me.

I tried to test it on qemu virt machine but it blows up but I figured out that the problem is it tries to copy the vector table over the top of the code itself which is loaded at 0x0. I'll look into moving the load address to a higher spot, either by linking it differently and hoping -kernel honors that, or having it relocate itself to the final location in the top of start.S so that the vector table is free to be at 0. Sadly due to the way the first two vectors work on 68k you can't *technically* use them for a branch instruction to the real code (though in practice you really could for the qemu virt machine).

Anyway, taking this as-is. Thanks!

PR #502, opened 2026-04-13
[lib][libc] Fix fclose condition for fs-backed streams

bl4ckhyun:
`fopen()` marks filesystem-backed streams with `use_fs = true`, but `fclose()`
checks the inverse condition before cleanup.

Fix the condition so `fclose()` closes and frees FS-backed streams correctly.


travisg:
this got picked up with the later PR, thanks.

PR #503, opened 2026-04-13
[lib][fs] Fix cleanup on handle allocation failure

bl4ckhyun:
When `fs_open_file()`, `fs_create_file()`, or `fs_open_dir()` fail to allocate
their wrapper handle objects, they should release the already-opened file or
directory cookie before dropping the mount reference.

Also return `ERR_NO_MEMORY` from `fs_create_file()` on allocation failure.


travisg:
this got picked up with the later PR.

PR #504, opened 2026-04-13
[platform][zynq] Check for CFI buffer allocation failure

bl4ckhyun:
Check the CFI buffer allocation result in `spiflash_detect()` before using it.

Return `ERR_NO_MEMORY` when the temporary buffer cannot be allocated.


travisg (approved):

travisg:
good catches, LGTM.

PR #505, opened 2026-04-15
[lib][fdtwalk] fix unaligned FDT memory reservation

bl4ckhyun:
Reserve the full page range that covers the FDT byte range.

The previous calculation rounded the base up and the size up separately,
which could miss the first page when the FDT physical address was not
page aligned.

Fix this by rounding the base down to the containing page boundary and
including the resulting offset in the length calculation.


travisg (approved):
LGTM, thanks!

PR #506, opened 2026-04-25
[dev][virtio-rng] Introduce virtio-rng driver to seed system prng

visitorckw:
Introduces a new virtio-rng driver to fetch hardware entropy from the host in virtualized environments.

Additionally, an early boot hook is implemented to automatically fetch entropy and seed the system's prng via srand() after the device is initialized.

### DIFF
--- a/dev/virtio/rng/virtio-rng.cpp
+++ b/dev/virtio/rng/virtio-rng.cpp
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2026 Kuan-Wei Chiu <visitorckw@gmail.com>
+ *
+ * Use of this source code is governed by a MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT
+ */
+
+#include <dev/virtio/rng.h>
+#include <dev/virtio/virtio-device.h>
+#include <dev/virtio/virtio_ring.h>
+#include <kernel/thread.h>
+#include <kernel/vm.h>
+#include <lk/debug.h>
+#include <lk/err.h>
+#include <lk/init.h>
+#include <lk/trace.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define LOCAL_TRACE 0
+
+constexpr uint16_t RNG_QUEUE_INDEX = 0;
+
+struct virtio_rng_state {
+    virtio_device *dev;
+    bool initialized;
+    volatile size_t last_rx_len;
+};
+
+static virtio_rng_state rng;
+
+static enum handler_return virtio_rng_irq(virtio_device *dev, uint ring_index, const vring_used_elem *e) {
+    if (ring_index == RNG_QUEUE_INDEX && e)
+        rng.last_rx_len = e->len;
+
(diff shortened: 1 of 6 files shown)


dev/virtio/rng/virtio-rng.cpp
```diff
@@ -0,0 +1,109 @@
...
+    rng.initialized = true;
+
+    LTRACEF("virtio-rng: initialized successfully\n");
+    return NO_ERROR;
+}
+
+ssize_t virtio_rng_read(void *buf, size_t len) {
+    if (!rng.initialized || len == 0) return ERR_NOT_CONFIGURED;
+
+    paddr_t paddr = vaddr_to_paddr(buf);
```
travisg, line 70: There's technically an issue here if the buffer crosses a page boundary and would thus need potentially two page lookups, I'd be okay with just doing a page crossing test here and reject it if it crosses one. The code below should be fine, because it's only using a single aligned word off the stack which by definition cant cross a page boundary.

dev/virtio/virtio-pci-bus.cpp
```diff
@@ -465,17 +468,45 @@ static status_t init_gpu(pci_location_t loc, const virtio_pci_devices &dev_table
...
+#else
+    return ERR_NOT_FOUND;
+#endif
+}
+
 
 int virtio_pci_init() {
     LTRACE_ENTRY;
 
     constexpr virtio_pci_devices devices[] = {
```
travisg, line 501: can you also add a virtio-mmio interface for this? Should be pretty easy, can test with the arm64 qemu since it can run with both mmio and PCI virtio interfaces (i think the -C flag switches to PCI from the default mmio)

travisg (changes requested):
Looks great! just a few nits and this is lovely. Thanks!

dev/virtio/rng/virtio-rng.cpp
```diff
@@ -0,0 +1,109 @@
...
+    unsigned int seed = 0;
+    ssize_t bytes_read;
+
+    if (!rng.initialized)
+        return;
+
+    bytes_read = virtio_rng_read(&seed, sizeof(seed));
+    
+    if (bytes_read == sizeof(seed)) {
+        srand(seed);
```
travisg, line 107: reminds me we should implement a better system RNG

dev/virtio/rng/virtio-rng.cpp
visitorckw: Sorry for the late reply.
Good catch! I overlooked the page boundary case.
I'll update the code to check if the buffer crosses a page boundary.

dev/virtio/virtio-pci-bus.cpp
visitorckw: Ack. Will do.

dev/virtio/rng/virtio-rng.cpp
visitorckw: I would be glad to help improve the system rng.
If there is anything specific you need help with or any direction you'd like me to explore, just let me know!

PR #507, opened 2026-05-06
[fs][fat] fix false EOF error on cluster-aligned reads

leon-chen1127:
When a file read reaches the end of a cluster and the total read length is exactly aligned with the cluster boundary, the iterator speculatively fetches the next sector. This causes the FAT driver to hit the EOF marker in the FAT table, returning ERR_OUT_OF_RANGE.

This patch adds a check to skip the next sector fetch if the target length has already been met.

Tested with a 4KB/8KB file on a FAT12/FAT16/FAT32 image for typical cluster sizes (e.g., 1KB, 2KB, 4KB), ensuring that the cluster boundary condition is triggered regardless of the specific FAT type.


travisg:
Great catch! I'm about to roll a big pile of changes (read-write support) to master from the wip/fat branch but it looks like this would apply on top of it cleanly so I'll hold off merging this in until I get that branch in, which should happen soon.

Thanks!

travisg:
@copilot resolve the merge conflicts in this pull request

travisg:
Resolved and pushed upstream. It didn't pick up that it completed this pull, but it did. Thanks!

PR #508, opened 2026-05-06
[ci] scripts: rewrite unittest to use expect

zhangxp1998:
Rewrite the python-based unittest script into two distinct expect scripts: one for testing UEFI integration and another for executing general unittests. This allows FAT tests to accurately evaluate hardware dependencies by keeping virtio block devices out of the general unittest environment. Updates GitHub CI configurations to depend on the 'expect' utility and invoke both shell scripts instead of python.


muhammed9990hamza-alt (approved):

travisg:
Okay, looks okay to me. Thanks!

PR #511, opened 2026-05-18
[platform] Amiga port (m68k)

jrac:
Tested on real hardware and under emulation; combinations of accelerated and stock A500, A600, and A1200. All CPU models are supported, although 060 requires being built with 'M68K_CPU = 68060'. Current default target is 68000, for greatest compatibility.

Features and hardware support:

   - Basic 1bpp display driver
   - Keyboard driver
   - Serial console support
   - Timer hardware driver
   - Custom bootloader
   - Heap arena size is based on detected chip RAM size

Current limitations:

   - Floppy boot only, no IDE/ATA (tested with Gotek, should work with real disk)
   - Only chip RAM supported. Boots with Fast Mem present, but doesn't utilise it.
   - Basic graphical abilities, room for improvements in acceleration and colour depth.
   - Early kernel boot output only present on serial console, though I believe this is due to the nature of the current gfxconsole design.

### DIFF
--- a/platform/amiga/adfgen.c
+++ b/platform/amiga/adfgen.c
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2026 Josh Cummings
+ *
+ * Use of this source code is governed by a MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT
+ */
+
+#include <stdio.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define BOOTBLOCK_SIZE 1024
+#define ADF_SIZE       901120
+
+static uint32_t read_be(const unsigned char *p) {
+    return ((uint32_t)p[0] << 24) | ((uint32_t)p[1] << 16) | ((uint32_t)p[2] << 8) |
+           ((uint32_t)p[3] << 0);
+}
+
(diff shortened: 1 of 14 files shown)


jrac:
Updated force push is from a few files that were missed.

travisg:
Looks pretty good! Will need to figure out what to do about xdftool though. I dunno how it passed the CI since it seems like it should have failed when building the m68k targets...

travisg:
Oh interesting, with script/buildall it'll simply skip it because of the $(error) in the platform file. An unintended consequence, but actually accomplishes the task but only when selecting an architecture. If regular buildall it will still fail.

jrac:
I've put together some code that can generate a valid ADF header and checksum. I'll update the PR once I've tested some more and integrated it in to the build process.

This way we don't have to mess around with potentially pulling in Python and things in the CI runners.

travisg:
Thank you so much! Writing it in python is fine, there are other ports that require python3, but not any exotic libs necessarily.

jrac:
Added the ADF generation code and makefile changes in a separate commit.

Also pushed a couple of changes to the original commit that addresses some bootloader issues that were discovered during recent testing. This also included the missing shell app in project/amiga.mk.

travisg:
Looks good! Finally got to it but this looks nice and clean. Merging momentarily.

PR #512, opened 2026-06-16
[lib][uefi] implement missing protocols required by GBL

zhangxp1998:
## Summary

- add UEFI protocol implementations needed by GBL
- add boot buffer allocation support for UEFI boot flows
- print readable protocol GUID names in UEFI boot service traces

## Test

- `source .env && make qemu-virt-arm64-test`
- `git diff --check origin/master..HEAD`

### DIFF
--- a/lib/uefi/blockio_protocols.cpp
+++ b/lib/uefi/blockio_protocols.cpp
@@ -103,21 +103,38 @@ __WEAK EfiStatus open_block_device(EfiHandle handle, const void** intf) {
   return EFI_STATUS_SUCCESS;
 }
 
-EfiStatus list_block_devices(size_t *num_handles, EfiHandle **buf) {
+EfiStatus list_block_devices(size_t *num_handles, EfiHandle *buf) {
+  if (num_handles == nullptr) {
+    return EFI_STATUS_INVALID_PARAMETER;
+  }
+
   size_t device_count = 0;
   bio_iter_devices([&device_count](bdev_t *dev) {
     device_count++;
     return true;
   });
-  auto devices =
-      reinterpret_cast<EfiHandle *>(uefi_malloc(sizeof(EfiHandle) * device_count));
+
+  if (device_count == 0) {
+    *num_handles = 0;
+    return EFI_STATUS_NOT_FOUND;
+  }
+
+  if (*num_handles < device_count) {
+    *num_handles = device_count;
+    return EFI_STATUS_BUFFER_TOO_SMALL;
+  }
+
+  // Allow null buffers for size queries, but require storage before writing.
+  if (buf == nullptr) {
+    return EFI_STATUS_INVALID_PARAMETER;
+  }
+
   size_t i = 0;
-  bio_iter_devices([&i, devices, device_count](bdev_t *dev) {
-    devices[i] = dev->name;
+  bio_iter_devices([&i, buf, device_count](bdev_t *dev) {
+    buf[i] = dev->name;
     i++;
     return i < device_count;
   });
   *num_handles = i;
-  *buf = devices;
   return EFI_STATUS_SUCCESS;
 }
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -16,6 +16,7 @@
  */
 #include "boot_service_provider.h"
 
+#include <endian.h>
 #include <lk/compiler.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -39,6 +40,63 @@
 
 namespace {
 
+bool guid_eq(const EfiGuid *a, const EfiGuid *b) {
+  if (a == nullptr || b == nullptr) {
+    return false;
+  }
+  return memcmp(a, b, sizeof(*a)) == 0;
+}
+
+bool guid_eq(const EfiGuid *a, const EfiGuid &b) {
+  return guid_eq(a, &b);
+}
+
+const char *protocol_name(const EfiGuid *protocol) {
+  if (protocol == nullptr) {
+    return nullptr;
+  }
+
+  struct ProtocolGuidName {
+    const EfiGuid *guid;
+    const char *name;
+  };
+  static constexpr ProtocolGuidName kProtocolGuidNames[] = {
+      {&LOADED_IMAGE_PROTOCOL_GUID, "LOADED_IMAGE_PROTOCOL_GUID"},
+      {&EFI_DEVICE_PATH_PROTOCOL_GUID, "EFI_DEVICE_PATH_PROTOCOL_GUID"},
+      {&LINUX_EFI_LOADED_IMAGE_FIXED_GUID,
+       "LINUX_EFI_LOADED_IMAGE_FIXED_GUID"},
+      {&EFI_RNG_PROTOCOL_GUID, "EFI_RNG_PROTOCOL_GUID"},
+      {&EFI_TCG2_PROTOCOL_GUID, "EFI_TCG2_PROTOCOL_GUID"},
+      {&EFI_LOAD_FILE2_PROTOCOL_GUID, "EFI_LOAD_FILE2_PROTOCOL_GUID"},
+      {&EFI_BLOCK_IO_PROTOCOL_GUID, "EFI_BLOCK_IO_PROTOCOL_GUID"},
+      {&EFI_BLOCK_IO2_PROTOCOL_GUID, "EFI_BLOCK_IO2_PROTOCOL_GUID"},
+      {&EFI_TEXT_INPUT_PROTOCOL_GUID, "EFI_TEXT_INPUT_PROTOCOL_GUID"},
+      {&EFI_GBL_OS_CONFIGURATION_PROTOCOL_GUID,
+       "EFI_GBL_OS_CONFIGURATION_PROTOCOL_GUID"},
+      {&EFI_GBL_EFI_IMAGE_LOADING_PROTOCOL_GUID,
+       "EFI_GBL_EFI_IMAGE_LOADING_PROTOCOL_GUID"},
+      {&EFI_GBL_EFI_BOOT_CONTROL_PROTOCOL_GUID,
+       "EFI_GBL_EFI_BOOT_CONTROL_PROTOCOL_GUID"},
+      {&EFI_GBL_EFI_AVB_PROTOCOL_GUID, "EFI_GBL_EFI_AVB_PROTOCOL_GUID"},
+      {&EFI_GBL_EFI_FASTBOOT_PROTOCOL_GUID,
+       "EFI_GBL_EFI_FASTBOOT_PROTOCOL_GUID"},
+      {&EFI_GBL_EFI_FASTBOOT_TRANSPORT_PROTOCOL_GUID,
+       "EFI_GBL_EFI_FASTBOOT_TRANSPORT_PROTOCOL_GUID"},
+      {&EFI_DT_FIXUP_PROTOCOL_GUID, "EFI_DT_FIXUP_PROTOCOL_GUID"},
+      {&EFI_TIMESTAMP_PROTOCOL_GUID, "EFI_TIMESTAMP_PROTOCOL_GUID"},
+      {&EFI_BOOT_MEMORY_PROTOCOL_GUID, "EFI_BOOT_MEMORY_PROTOCOL_GUID"},
+      {&EFI_ERASE_BLOCK_PROTOCOL_GUID, "EFI_ERASE_BLOCK_PROTOCOL_GUID"},
+  };
+
+  for (const auto &entry : kProtocolGuidNames) {
+    if (guid_eq(protocol, entry.guid)) {
+      return entry.name;
+    }
+  }
+
+  return nullptr;
+}
+
 EfiStatus open_protocol(EfiHandle handle, const EfiGuid *protocol, const void **intf,
                         EfiHandle agent_handle, EfiHandle controller_handle,
                         EfiOpenProtocolAttributes attr);
@@ -47,41 +105,56 @@ EfiStatus locate_handle_buffer(EfiLocateHandleSearchType search_type,
                                const EfiGuid *protocol, const void *search_key,
                                size_t *num_handles, EfiHandle **buf);
 
+uint64_t guid_data4(const EfiGuid *guid) {
+  uint64_t data4 = 0;
+  memcpy(&data4, guid->data4, sizeof(data4));
+  return LE64(data4);
+}
+
+const char *fmt_guid(const EfiGuid *protocol, char *buf, size_t buf_size) {
+  const char *name = protocol_name(protocol);
+  if (name != nullptr) {
+    return name;
+  }
+  if (protocol == nullptr) {
+    return "nullptr";
+  }
+  snprintf(buf, buf_size, "(0x%x 0x%x 0x%x 0x%llx)", protocol->data1,
+           protocol->data2, protocol->data3,
+           static_cast<unsigned long long>(guid_data4(protocol)));
+  return buf;
+}
+
 EfiStatus unload(EfiHandle handle) { return EFI_STATUS_SUCCESS; }
 
-bool guid_eq(const EfiGuid *a, const EfiGuid *b) {
-  return memcmp(a, b, sizeof(*a)) == 0;
-}
-
-bool guid_eq(const EfiGuid *a, const EfiGuid &b) {
-  return memcmp(a, &b, sizeof(*a)) == 0;
-}
-
 EfiHandle singleton_protocol_handle(const EfiGuid *protocol) {
-  if (guid_eq(protocol, EFI_GBL_OS_CONFIGURATION_PROTOCOL_GUID)) {
-    return &EFI_GBL_OS_CONFIGURATION_PROTOCOL_GUID;
-  } else if (guid_eq(protocol, EFI_DT_FIXUP_PROTOCOL_GUID)) {
-    return &EFI_DT_FIXUP_PROTOCOL_GUID;
-  } else if (guid_eq(protocol, EFI_TIMESTAMP_PROTOCOL_GUID)) {
-    return &EFI_TIMESTAMP_PROTOCOL_GUID;
-  } else if (guid_eq(protocol, EFI_BOOT_MEMORY_PROTOCOL_GUID)) {
-    return &EFI_BOOT_MEMORY_PROTOCOL_GUID;
-  } else if (guid_eq(protocol, EFI_GBL_EFI_AVB_PROTOCOL_GUID)) {
-    return &EFI_GBL_EFI_AVB_PROTOCOL_GUID;
-  } else if (guid_eq(protocol, EFI_GBL_EFI_BOOT_CONTROL_PROTOCOL_GUID)) {
-    return &EFI_GBL_EFI_BOOT_CONTROL_PROTOCOL_GUID;
-  } else if (guid_eq(protocol, EFI_RNG_PROTOCOL_GUID)) {
-    return &EFI_RNG_PROTOCOL_GUID;
+  static constexpr const EfiGuid *kSingletonGuids[] = {
+      &EFI_GBL_OS_CONFIGURATION_PROTOCOL_GUID,
+      &EFI_DT_FIXUP_PROTOCOL_GUID,
+      &EFI_TIMESTAMP_PROTOCOL_GUID,
+      &EFI_BOOT_MEMORY_PROTOCOL_GUID,
+      &EFI_GBL_EFI_AVB_PROTOCOL_GUID,
+      &EFI_GBL_EFI_BOOT_CONTROL_PROTOCOL_GUID,
+      &EFI_RNG_PROTOCOL_GUID,
+  };
+
+  for (const EfiGuid *singleton_guid : kSingletonGuids) {
+    if (guid_eq(protocol, singleton_guid)) {
+      return singleton_guid;
+    }
   }
   return nullptr;
 }
 
 EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
                           void **intf) {
+  char name_buf[64];
+  printf("%s(%p, %s, %p);\n", __FUNCTION__, handle,
+         fmt_guid(protocol, name_buf, sizeof(name_buf)), intf);
+
   if (guid_eq(protocol, LOADED_IMAGE_PROTOCOL_GUID)) {
-    printf("handle_protocol(%p, LOADED_IMAGE_PROTOCOL_GUID, %p);\n", handle,
-           intf);
-    const auto loaded_image = static_cast<EfiLoadedImageProtocol *>(uefi_malloc(sizeof(EfiLoadedImageProtocol)));
+    const auto loaded_image = static_cast<EfiLoadedImageProtocol *>(
+        uefi_malloc(sizeof(EfiLoadedImageProtocol)));
     if (!loaded_image) {
       return EFI_STATUS_OUT_OF_RESOURCES;
     }
@@ -97,18 +170,16 @@ EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
     *intf = loaded_image;
     return EFI_STATUS_SUCCESS;
   } else if (guid_eq(protocol, LINUX_EFI_LOADED_IMAGE_FIXED_GUID)) {
-    printf("handle_protocol(%p, LINUX_EFI_LOADED_IMAGE_FIXED_GUID, %p);\n",
-           handle, intf);
     return EFI_STATUS_UNSUPPORTED;
-  } else {
-    printf("handle_protocol(%p, %p, %p);\n", handle, protocol, intf);
   }
   return EFI_STATUS_UNSUPPORTED;
 }
 
 EfiStatus register_protocol_notify(const EfiGuid *protocol, EfiEvent event,
                                    void **registration) {
-  printf("%s is unsupported\n", __FUNCTION__);
+  char name_buf[64];
+  printf("%s(%s, event=%p, registration=%p) is unsupported\n", __FUNCTION__,
+         fmt_guid(protocol, name_buf, sizeof(name_buf)), event, registration);
   return EFI_STATUS_UNSUPPORTED;
 }
 
@@ -126,20 +197,18 @@ EfiStatus locate_handle(EfiLocateHandleSearchType search_type,
       *buf_size = num_handles * sizeof(EfiHandle);
       return status;
     }
-    printf("%s(0x%x, EFI_BLOCK_IO_PROTOCOL_GUID, search_key=%p)\n",
-           __FUNCTION__, search_type, search_key);
+    char name_buf[64];
+    printf("%s(0x%x, %s, search_key=%p)\n", __FUNCTION__, search_type,
+           fmt_guid(protocol, name_buf, sizeof(name_buf)), search_key);
     return EFI_STATUS_UNSUPPORTED;
-  } else if (guid_eq(protocol, EFI_TEXT_INPUT_PROTOCOL_GUID)) {
-    printf("%s(0x%x, EFI_TEXT_INPUT_PROTOCOL_GUID, search_key=%p)\n",
-           __FUNCTION__, search_type, search_key);
-    return EFI_STATUS_NOT_FOUND;
   }
 
   const EfiHandle singleton_handle = singleton_protocol_handle(protocol);
-  if (singleton_handle != nullptr) {
-    printf("%s(0x%x, %p, search_key=%p)\n", __FUNCTION__, search_type, protocol,
-           search_key);
+  char name_buf[64];
+  printf("%s(0x%x, %s, search_key=%p)\n", __FUNCTION__, search_type,
+         fmt_guid(protocol, name_buf, sizeof(name_buf)), search_key);
 
+  if (singleton_handle != nullptr) {
     const size_t required_size = sizeof(EfiHandle);
     if (*buf_size < required_size) {
       *buf_size = required_size;
@@ -147,15 +216,13 @@ EfiStatus locate_handle(EfiLocateHandleSearchType search_type,
     }
 
     *buf_size = required_size;
-    if (buf != nullptr) {
-      *buf = singleton_handle;
+    if (buf == nullptr) {
+      return EFI_STATUS_INVALID_PARAMETER;
     }
+    *buf = singleton_handle;
     return EFI_STATUS_SUCCESS;
   }
 
-  printf("%s(0x%x, (0x%x 0x%x 0x%x 0x%llx), search_key=%p)\n", __FUNCTION__,
-         search_type, protocol->data1, protocol->data2, protocol->data3,
-         *(uint64_t *)&protocol->data4, search_key);
   return EFI_STATUS_NOT_FOUND;
 }
 
@@ -171,18 +238,19 @@ EfiStatus locate_protocol(const EfiGuid *protocol, void *registration,
                          nullptr, nullptr, 0);
   }
 
-  if (memcmp(protocol, &EFI_RNG_PROTOCOL_GUID, sizeof(*protocol)) == 0) {
-    printf("%s(EFI_RNG_PROTOCOL_GUID) is unsupported.\n", __FUNCTION__);
+  char name_buf[64];
+  const char *name = fmt_guid(protocol, name_buf, sizeof(name_buf));
+
+  if (guid_eq(protocol, EFI_RNG_PROTOCOL_GUID)) {
+    printf("%s(%s) is unsupported.\n", __FUNCTION__, name);
     return EFI_STATUS_UNSUPPORTED;
   }
-  if (memcmp(protocol, &EFI_TCG2_PROTOCOL_GUID, sizeof(*protocol)) == 0) {
-    printf("%s(EFI_TCG2_PROTOCOL_GUID) is unsupported.\n", __FUNCTION__);
+  if (guid_eq(protocol, EFI_TCG2_PROTOCOL_GUID)) {
+    printf("%s(%s) is unsupported.\n", __FUNCTION__, name);
     return EFI_STATUS_NOT_FOUND;
   }
 
-  printf("%s(%x %x %x %llx) is unsupported\n", __FUNCTION__, protocol->data1,
-         protocol->data2, protocol->data3,
-         *reinterpret_cast<const uint64_t *>(&protocol->data4));
+  printf("%s(%s) is unsupported\n", __FUNCTION__, name);
   return EFI_STATUS_NOT_FOUND;
 }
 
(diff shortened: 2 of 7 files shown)


lib/uefi/uefi_platform.cpp
```diff
@@ -202,6 +208,138 @@ EfiStatus erase_blocks(EfiEraseBlockProtocol* self, uint32_t media_id,
...
+}
+
+EfiStatus rng_get_rng(EfiRngProtocol* self, const EfiRngAlgorithm* rng_algorithm,
+                      size_t rng_value_length, uint8_t* rng_value) {
+  if (rng_value == nullptr) {
+    return EFI_STATUS_INVALID_PARAMETER;
+  }
+  memset(rng_value, 0, rng_value_length);
+  return EFI_STATUS_SUCCESS;
+}
```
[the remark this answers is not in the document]
zhangxp1998: Fixed in `1671f541`: `get_info` and `get_rng` now return `EFI_STATUS_UNSUPPORTED` instead of advertising success or returning non-random bytes.

lib/uefi/blockio_protocols.cpp
```diff
@@ -103,21 +103,24 @@ __WEAK EfiStatus open_block_device(EfiHandle handle, const void** intf) {
...
     device_count++;
     return true;
   });
-  auto devices =
-      reinterpret_cast<EfiHandle *>(uefi_malloc(sizeof(EfiHandle) * device_count));
+
+  if (*num_handles < device_count) {
+    *num_handles = device_count;
+    return EFI_STATUS_BUFFER_TOO_SMALL;
+  }
```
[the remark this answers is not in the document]
zhangxp1998: Fixed in `1671f541`: `list_block_devices()` now validates `num_handles`, returns `EFI_STATUS_NOT_FOUND` when no block devices exist, and validates `buf` before writing handles.

lib/uefi/boot_service_provider.cpp
```diff
@@ -167,36 +348,37 @@ void restore_tpl(EfiTpl old_tpl) {
 EfiStatus open_protocol(EfiHandle handle, const EfiGuid *protocol, const void **intf,
                         EfiHandle agent_handle, EfiHandle controller_handle,
                         EfiOpenProtocolAttributes attr) {
+  const char *name = protocol_name(protocol);
   if (guid_eq(protocol, LOADED_IMAGE_PROTOCOL_GUID)) {
     auto interface = reinterpret_cast<EfiLoadedImageProtocol *>(
         uefi_malloc(sizeof(EfiLoadedImageProtocol)));
     memset(interface, 0, sizeof(*interface));
     interface->parent_handle = handle;
```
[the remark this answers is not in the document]
zhangxp1998: Fixed in `eb17ff01`: the LOADED_IMAGE branch now checks `uefi_malloc()` and returns `EFI_STATUS_OUT_OF_RESOURCES` before `memset` if allocation fails.

lib/uefi/uefi_platform.cpp
```diff
@@ -94,7 +97,10 @@ __WEAK EfiStatus select_device_trees(struct GblEfiOsConfigurationProtocol* self,
                                      GblEfiVerifiedDeviceTree* device_trees) {
   printf("%s(%p, %p %lu)\n", __FUNCTION__, self, device_trees,
          num_device_trees);
-  return EFI_STATUS_UNSUPPORTED;
+  if (num_device_trees > 0) {
+    device_trees[0].selected = true;
+  }
+  return EFI_STATUS_SUCCESS;
```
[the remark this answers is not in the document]
zhangxp1998: Fixed in `1671f541`: `select_device_trees()` now returns `EFI_STATUS_INVALID_PARAMETER` for a null `device_trees` pointer when entries are present, and clears all `selected` flags before selecting the first tree.

lib/uefi/blockio_protocols.cpp
```diff
@@ -103,21 +103,37 @@ __WEAK EfiStatus open_block_device(EfiHandle handle, const void** intf) {
   return EFI_STATUS_SUCCESS;
 }
 
-EfiStatus list_block_devices(size_t *num_handles, EfiHandle **buf) {
+EfiStatus list_block_devices(size_t *num_handles, EfiHandle *buf) {
+  if (num_handles == nullptr) {
```
silverneko, line 107: why don't we also check `buf == nullptr` here?

lib/uefi/boot_service_provider.cpp
```diff
@@ -37,8 +37,68 @@
...
+      {&EFI_GBL_EFI_FASTBOOT_TRANSPORT_PROTOCOL_GUID,
+       "EFI_GBL_EFI_FASTBOOT_TRANSPORT_PROTOCOL_GUID"},
+      {&EFI_DT_FIXUP_PROTOCOL_GUID, "EFI_DT_FIXUP_PROTOCOL_GUID"},
+      {&EFI_TIMESTAMP_PROTOCOL_GUID, "EFI_TIMESTAMP_PROTOCOL_GUID"},
+      {&EFI_BOOT_MEMORY_PROTOCOL_GUID, "EFI_BOOT_MEMORY_PROTOCOL_GUID"},
+      {&EFI_ERASE_BLOCK_PROTOCOL_GUID, "EFI_ERASE_BLOCK_PROTOCOL_GUID"},
+  };
+
+  for (const auto &entry : kProtocolGuidNames) {
+    if (memcmp(protocol, entry.guid, sizeof(*protocol)) == 0) {
```
silverneko: compare EfiGuid with `guid_eq`

lib/uefi/boot_service_provider.cpp
```diff
@@ -70,47 +149,128 @@ EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
...
+           search_key);
+  } else if (protocol != nullptr) {
+    printf("%s(0x%x, (0x%x 0x%x 0x%x 0x%llx), search_key=%p)\n",
+           __FUNCTION__, search_type, protocol->data1, protocol->data2,
+           protocol->data3,
+           static_cast<unsigned long long>(guid_data4(protocol)), search_key);
+  } else {
+    printf("%s(0x%x, nullptr, search_key=%p)\n", __FUNCTION__, search_type,
+           search_key);
+  }
```
silverneko: nit: this same pattern repeated many times, could we factor this out?

How about make a function that converts `*EfiGuid` to `*char`,
`name` if a name is found,
`(0x%x 0x%x 0x%x 0x%llx)` if no name is found and non-null
`(nullptr)` if null

lib/uefi/boot_service_provider.cpp
```diff
@@ -49,11 +109,30 @@ bool guid_eq(const EfiGuid *a, const EfiGuid &b) {
...
+  return nullptr;
+}
+
 EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
                           void **intf) {
+  const char *name = protocol_name(protocol);
   if (guid_eq(protocol, LOADED_IMAGE_PROTOCOL_GUID)) {
-    printf("handle_protocol(%p, LOADED_IMAGE_PROTOCOL_GUID, %p);\n", handle,
-           intf);
+    printf("%s(%p, %s, %p);\n", __FUNCTION__, handle, name, intf);
```
silverneko: What if `name` is nullptr? how do we ensure that `protocol_name` always return non-null for all know GUIDs?

lib/uefi/boot_service_provider.cpp
zhangxp1998: Good point. I avoided adding a helper that converts `EfiGuid*` to `char*` because this code runs in a freestanding LK environment and does not use C++ conveniences like `std::string`. A helper that returns formatted GUID text would need caller-owned storage or static storage, which felt heavier than keeping the formatting at the logging call sites for now.

So I kept `protocol_name()` as a lightweight known-GUID lookup, and unknown/non-null GUIDs are still formatted numerically by each caller.

lib/uefi/boot_service_provider.cpp
zhangxp1998: `protocol_name()` is intentionally best-effort: we should keep its table up to date with all known GUIDs as we add protocol support, but it can still return `nullptr` for unknown GUIDs. Callers should treat `nullptr` as part of the contract and handle it explicitly, either by printing `nullptr` for a null protocol pointer or by formatting the GUID fields numerically for an unknown non-null GUID.

I also updated the GUID comparisons to use `guid_eq()`, and made `guid_eq()` return false for null pointers so these checks are safer.

lib/uefi/blockio_protocols.cpp
zhangxp1998: Done.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done.

lib/uefi/boot_service_provider.cpp
silverneko: I agree, but you didn't check the null-ness of `name` here? I'm suggesting that we should not assume `name` is non-null, and we explicitly check it before passing it to printf, furthermore, we can factor out the common "printf" call in each if-arm to the beginning of the function:

```cpp
EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
                          void **intf) {
  const char *name = protocol_name(protocol);
  if (name) {
    printf("%s(%p, %s, %p);\n", __FUNCTION__, handle, name, intf);
  } else if (protocol) {
    printf("%s(%p, (0x%x 0x%x 0x%x 0x%llx), %p);\n", __FUNCTION__, handle,
           protocol->data1, protocol->data2, protocol->data3,
           static_cast<unsigned long long>(guid_data4(protocol)), intf);
  } else {
    printf("%s(%p, nullptr, %p);\n", __FUNCTION__, handle, intf);
  }

  if (guid_eq(protocol, LOADED_IMAGE_PROTOCOL_GUID)) {
    const auto loaded_image = static_cast<EfiLoadedImageProtocol *>(uefi_malloc(sizeof(EfiLoadedImageProtocol)));
    if (!loaded_image) {
      return EFI_STATUS_OUT_OF_RESOURCES;
    }
    *loaded_image = {};
    loaded_image->revision = EFI_LOADED_IMAGE_PROTOCOL_REVISION;
    loaded_image->parent_handle = nullptr;
    loaded_image->system_table = nullptr;
    loaded_image->load_options_size = 0;
    loaded_image->load_options = nullptr;
    loaded_image->unload = unload;
    loaded_image->image_base = handle;

    *intf = loaded_image;
    return EFI_STATUS_SUCCESS;
  } else if (guid_eq(protocol, LINUX_EFI_LOADED_IMAGE_FIXED_GUID)) {
    return EFI_STATUS_UNSUPPORTED;
  }
  return EFI_STATUS_UNSUPPORTED;
}
```

lib/uefi/boot_service_provider.cpp
silverneko: I agree 100% using std::string would be too heavy. I was thinking something like `snprintf` on a stack-allocated buffer. This way the memory footprint would be predictable, and we only incur slight overhead (a single snprintf call when `name` is NULL).

```cpp
// fmt_guid could be function or macro?
const char* fmt_guid(const EfiGuid *guid, char *buf /* must be a stack buffer or stack-local storage, at least sizeof (0x%x 0x%x 0x%x 0x%llx) */) {
  if (!name || !protocol) return "(nullptr)";
  if (const char* name = protocol_name(guid)) return name;
  snprintf(buf, /* sizeof buf */, "(0x%x 0x%x 0x%x 0x%llx)",
           search_type, protocol->data1, protocol->data2,
           protocol->data3,
           static_cast<unsigned long long>(guid_data4(protocol)));
  return buf;
}
```

caller:
```cpp
char buf[100] = "";
const char *name = fmt_guid(protocol, buf);
printf("%s(0x%x, %s, search_key=%p)\n", __FUNCTION__, search_type, name,
           search_key);
```

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in `03100e76`: `handle_protocol()` now formats the protocol name once at entry, so unknown GUIDs and null protocols are handled before the common `printf()` call.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in `03100e76`: added a stack-buffer `fmt_guid()` helper using `snprintf()` for unknown GUIDs, and reused it in `handle_protocol()`, `register_protocol_notify()`, and `locate_handle()`.

lib/uefi/boot_service_provider.cpp
```diff
@@ -39,22 +39,130 @@
...
+  return static_cast<uint64_t>(guid->data4[0]) |
+         (static_cast<uint64_t>(guid->data4[1]) << 8) |
+         (static_cast<uint64_t>(guid->data4[2]) << 16) |
+         (static_cast<uint64_t>(guid->data4[3]) << 24) |
+         (static_cast<uint64_t>(guid->data4[4]) << 32) |
+         (static_cast<uint64_t>(guid->data4[5]) << 40) |
+         (static_cast<uint64_t>(guid->data4[6]) << 48) |
+         (static_cast<uint64_t>(guid->data4[7]) << 56);
+}
+
```
silverneko: optional: you can use builtin macros to do endian conversions https://github.com/littlekernel/lk/blob/1e769e006b42981f580f4b9492f20d8a70525de9/lib/libc/include/endian.h#L36

```
uint64_t data4 = 0;
memcpy(&data4, guid->data4, sizeof(data4));
data4 = LE64(data4);
return data4;
```

lib/uefi/boot_service_provider.cpp
```diff
@@ -39,22 +39,130 @@
...
+    return &EFI_TIMESTAMP_PROTOCOL_GUID;
+  } else if (guid_eq(protocol, EFI_BOOT_MEMORY_PROTOCOL_GUID)) {
+    return &EFI_BOOT_MEMORY_PROTOCOL_GUID;
+  } else if (guid_eq(protocol, EFI_GBL_EFI_AVB_PROTOCOL_GUID)) {
+    return &EFI_GBL_EFI_AVB_PROTOCOL_GUID;
+  } else if (guid_eq(protocol, EFI_GBL_EFI_BOOT_CONTROL_PROTOCOL_GUID)) {
+    return &EFI_GBL_EFI_BOOT_CONTROL_PROTOCOL_GUID;
+  } else if (guid_eq(protocol, EFI_RNG_PROTOCOL_GUID)) {
+    return &EFI_RNG_PROTOCOL_GUID;
+  }
```
silverneko: Use a constant list of singleton GUIDs to simplify long if-else to a single loop:

```suggestion
  constexpr EfiGuid *kSingletonGuids[] = {&EFI_GBL_OS_CONFIGURATION_PROTOCOL_GUID, ...};
  for (const auto g : kSingletonGuids) {
    if (guid_eq(protocol, g)) {
      return g
    }
  }
```

lib/uefi/boot_service_provider.cpp
```diff
@@ -70,47 +178,99 @@ EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
...
+    if (buf == nullptr) {
+      return EFI_STATUS_INVALID_PARAMETER;
+    }
+    *buf = singleton_handle;
+    return EFI_STATUS_SUCCESS;
+  }
+
+  printf("%s(0x%x, %s, search_key=%p)\n", __FUNCTION__, search_type,
+         fmt_guid(protocol, name_buf, sizeof(name_buf)), search_key);
+  return EFI_STATUS_NOT_FOUND;
```
silverneko:
```suggestion
  printf("%s(0x%x, %s, search_key=%p)\n", __FUNCTION__, search_type,
         fmt_guid(protocol, name_buf, sizeof(name_buf)), search_key);
  if (singleton_handle != nullptr) {
    const size_t required_size = sizeof(EfiHandle);
    if (*buf_size < required_size) {
      *buf_size = required_size;
      return EFI_STATUS_BUFFER_TOO_SMALL;
    }

    *buf_size = required_size;
    if (buf == nullptr) {
      return EFI_STATUS_INVALID_PARAMETER;
    }
    *buf = singleton_handle;
    return EFI_STATUS_SUCCESS;
  }

  return EFI_STATUS_NOT_FOUND;
```

lib/uefi/boot_service_provider.cpp
```diff
@@ -70,47 +178,99 @@ EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
...
+      return status;
+    }
+    printf("%s(0x%x, %s, search_key=%p)\n", __FUNCTION__, search_type,
+           fmt_guid(protocol, name_buf, sizeof(name_buf)), search_key);
+    return EFI_STATUS_UNSUPPORTED;
+  } else if (guid_eq(protocol, EFI_TEXT_INPUT_PROTOCOL_GUID)) {
+    printf("%s(0x%x, %s, search_key=%p)\n", __FUNCTION__, search_type,
+           fmt_guid(protocol, name_buf, sizeof(name_buf)), search_key);
+    return EFI_STATUS_NOT_FOUND;
+  }
```
silverneko: why do we need to specifically check `EFI_TEXT_INPUT_PROTOCOL_GUID` here?

`EFI_TEXT_INPUT_PROTOCOL_GUID` is already not a "singleton", so it would report `EFI_STATUS_NOT_FOUND` @L240 anyway?

lib/uefi/boot_service_provider.cpp
```diff
@@ -70,47 +178,99 @@ EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
...
-  return EFI_STATUS_UNSUPPORTED;
+  if (name != nullptr) {
+    printf("%s(%s) is unsupported\n", __FUNCTION__, name);
+  } else {
+    printf("%s(0x%x 0x%x 0x%x 0x%llx) is unsupported\n", __FUNCTION__,
+           protocol->data1, protocol->data2, protocol->data3,
+           static_cast<unsigned long long>(guid_data4(protocol)));
+  }
+  return EFI_STATUS_NOT_FOUND;
 }
```
silverneko, line 255: Format the protocol name with fmt_guid, otherwise `name` could be nullptr, and printf could result in null pointer exception.

simplify to:

```
  if (protocol == nullptr || intf == nullptr) {
    return EFI_STATUS_INVALID_PARAMETER;
  }

  const EfiHandle singleton_handle = singleton_protocol_handle(protocol);
  if (singleton_handle != nullptr) {
    return open_protocol(singleton_handle, protocol, const_cast<const void **>(intf),
                         nullptr, nullptr, 0);
  }

  char name_buf[64] = "";
  const char *name = fmt_guid(protocol, name_buf, sizeof(name_buf));
  printf("%s(%s) is unsupported.\n", __FUNCTION__, name);

  if (guid_eq(protocol, EFI_RNG_PROTOCOL_GUID)) {
    return EFI_STATUS_UNSUPPORTED;
  }
  return EFI_STATUS_NOT_FOUND;
```

lib/uefi/boot_service_provider.cpp
```diff
@@ -124,7 +284,19 @@ EfiStatus calculate_crc32(void *data, size_t len, uint32_t *crc32) {
...
+           handle, intf);
+  } else if (protocol != nullptr) {
+    printf("%s((0x%x 0x%x 0x%x 0x%llx), handle=%p, intf=%p) is "
+           "unsupported\n",
+           __FUNCTION__, protocol->data1, protocol->data2, protocol->data3,
+           static_cast<unsigned long long>(guid_data4(protocol)), handle, intf);
+  } else {
+    printf("%s(nullptr, handle=%p, intf=%p) is unsupported\n", __FUNCTION__,
+           handle, intf);
+  }
```
silverneko: should just call fmt_guid?

```
EfiStatus uninstall_protocol_interface(EfiHandle handle,
                                       const EfiGuid *protocol, void *intf) {
printf("%s(%s, handle=%p, intf=%p) is unsupported\n", __FUNCTION__, fmt_guid(protocol, ...),
           handle, intf);
return EFI_STATUS_UNSUPPORTED;
}
```

lib/uefi/boot_service_provider.cpp
```diff
@@ -137,11 +309,19 @@ EfiStatus load_image(bool boot_policy, EfiHandle parent_image_handle,
...
+  const char *name = protocol_name(protocol);
+  if (name != nullptr) {
+    printf("%s(%s) is unsupported\n", __FUNCTION__, name);
+  } else if (protocol != nullptr) {
+    printf("%s(0x%x 0x%x 0x%x 0x%llx) is unsupported\n", __FUNCTION__,
+           protocol->data1, protocol->data2, protocol->data3,
+           static_cast<unsigned long long>(guid_data4(protocol)));
+  } else {
+    printf("%s(nullptr) is unsupported\n", __FUNCTION__);
+  }
```
silverneko: same, just call fmt_guid to format the protocol name?

lib/uefi/boot_service_provider.cpp
```diff
@@ -246,113 +428,140 @@ EfiStatus open_protocol(EfiHandle handle, const EfiGuid *protocol, const void **
...
+  } else if (protocol != nullptr) {
+    printf("%s((0x%x 0x%x 0x%x 0x%llx), handle=%p, agent_handle=%p, "
+           "controller_handle=%p, attr=0x%x) is unsupported\n",
+           __FUNCTION__, protocol->data1, protocol->data2, protocol->data3,
+           static_cast<unsigned long long>(guid_data4(protocol)), handle,
+           agent_handle, controller_handle, attr);
+  } else {
+    printf("%s(nullptr, handle=%p, agent_handle=%p, controller_handle=%p, "
+           "attr=0x%x) is unsupported\n",
+           __FUNCTION__, handle, agent_handle, controller_handle, attr);
```
silverneko: fmt_guid

lib/uefi/boot_service_provider.cpp
```diff
@@ -246,113 +428,140 @@ EfiStatus open_protocol(EfiHandle handle, const EfiGuid *protocol, const void **
...
+           __FUNCTION__, name, handle, agent_handle, controller_handle, attr);
+    *intf = open_gbl_efi_boot_control_protocol();
+    if (*intf == nullptr) {
+      return EFI_STATUS_OUT_OF_RESOURCES;
+    }
+    return EFI_STATUS_SUCCESS;
+  } else if (guid_eq(protocol, EFI_RNG_PROTOCOL_GUID)) {
+    printf("%s(%s, handle=%p, agent_handle=%p, controller_handle=%p, "
+           "attr=0x%x)\n",
+           __FUNCTION__, name, handle, agent_handle, controller_handle, attr);
```
silverneko: Since almost all if-arm calls the same `printf` statement, what do you think if we just call printf once before entering the if-chain?

lib/uefi/boot_service_provider.cpp
```diff
@@ -246,113 +428,140 @@ EfiStatus open_protocol(EfiHandle handle, const EfiGuid *protocol, const void **
...
-  printf("%s is unsupported 0x%x 0x%x 0x%x 0x%llx\n", __FUNCTION__,
-         protocol->data1, protocol->data2, protocol->data3,
-         *(uint64_t *)&protocol->data4);
   return EFI_STATUS_UNSUPPORTED;
 }
 
 EfiStatus close_protocol(EfiHandle handle, const EfiGuid *protocol,
                          EfiHandle agent_handle, EfiHandle controller_handle) {
+  const char *name = protocol_name(protocol);
   if (guid_eq(protocol, LOADED_IMAGE_PROTOCOL_GUID)) {
```
silverneko, line 412: same.  what do you think if we just call printf once before entering the if-chain?
also use fmt_guid to format guids

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8. The remaining common trace sites now use fmt_guid() before the if-chain.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8. The stack-buffer fmt_guid() helper is now used by the remaining log sites.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8. locate_protocol() now uses fmt_guid() for the unsupported log path.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8. Switched this to fmt_guid().

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8. Moved the common trace before the if-chain.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in a9a4fac8. Moved the common trace before the if-chain and use fmt_guid().

lib/uefi/boot_service_provider.cpp
zhangxp1998: Ack, thanks. Keeping the explicit byte assembly here since this helper is only used for debug logging.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Ack. Leaving this as-is for now to keep this patch focused on the logging cleanup.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Ack. This patch keeps the existing locate_handle() control flow and only moves the remaining GUID formatting to fmt_guid().

lib/uefi/boot_service_provider.cpp
zhangxp1998: Good point. Kept this unchanged in this patch to avoid changing locate_handle() behavior beyond the logging cleanup.

lib/uefi/boot_service_provider.h
```diff
@@ -21,6 +21,7 @@
 #include <uefi/system_table.h>
 
 void setup_boot_service_table(EfiBootService *service);
+const char *protocol_name(const EfiGuid *protocol);
```
silverneko: How do you plan to use this function outside of boot_service_provider.cpp? It no, we can make this a private function / don't export.

lib/uefi/uefi_platform.cpp
```diff
@@ -94,7 +97,16 @@ __WEAK EfiStatus select_device_trees(struct GblEfiOsConfigurationProtocol* self,
...
          num_device_trees);
-  return EFI_STATUS_UNSUPPORTED;
+  if (num_device_trees > 0) {
+    if (device_trees == nullptr) {
+      return EFI_STATUS_INVALID_PARAMETER;
+    }
+    for (size_t i = 0; i < num_device_trees; i++) {
+      device_trees[i].selected = false;
+    }
+    device_trees[0].selected = true;
```
silverneko: https://android.googlesource.com/platform/bootable/libbootloader/+/refs/heads/gbl-mainline/gbl/docs/gbl_efi_os_configuration_protocol.md#gbl_efi_os_configuration_protocol_selectdevicetrees

says exactly one base device tree must be selected, and any number of DT overlay can be selected. Should we refine this to be `device_trees[i].selected = true;` where `i` is the smallest index whose `device_trees[i].type == GBL_EFI_DEVICE_TREE_TYPE_DEVICE_TREE`?

lib/uefi/uefi_platform.h
```diff
@@ -85,4 +85,10 @@ EfiStatus open_efi_erase_block_protocol(EfiHandle handle, const void** intf);
 
 GblEfiBootMemoryProtocol* open_boot_memory_protocol();
 
+struct GblEfiAvbProtocol* open_gbl_efi_avb_protocol();
+
+struct GblEfiBootControlProtocol* open_gbl_efi_boot_control_protocol();
+
+struct EfiRngProtocol* open_efi_rng_protocol();
```
silverneko: nit: the coding style is to omit the `struct` prefix and prefer the short name, i.e. just:

```
GblEfiAvbProtocol* open_gbl_efi_avb_protocol();

GblEfiBootControlProtocol* open_gbl_efi_boot_control_protocol();

EfiRngProtocol* open_efi_rng_protocol();
```

lib/uefi/boot_service_provider.cpp
```diff
@@ -70,47 +178,94 @@ EfiStatus handle_protocol(EfiHandle handle, const EfiGuid *protocol,
...
                           void **intf) {
-  if (protocol == nullptr) {
+  if (protocol == nullptr || intf == nullptr) {
     return EFI_STATUS_INVALID_PARAMETER;
   }
-  if (memcmp(protocol, &EFI_RNG_PROTOCOL_GUID, sizeof(*protocol)) == 0) {
-    printf("%s(EFI_RNG_PROTOCOL_GUID) is unsupported.\n", __FUNCTION__);
+
+  char name_buf[64];
+  const char *name = fmt_guid(protocol, name_buf, sizeof(name_buf));
```
silverneko: nit: move this after L256, as this is unused if we enter the "singleton_hnadle != nullptr" arm

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in `f8a1b66e`: `guid_data4()` now uses `memcpy()` plus `LE64()`.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in `f8a1b66e`: `singleton_protocol_handle()` now uses a singleton GUID list and a loop.

lib/uefi/boot_service_provider.h
zhangxp1998: Done in `f8a1b66e`: `protocol_name()` is now private to `boot_service_provider.cpp` and no longer exported in the header.

lib/uefi/uefi_platform.h
zhangxp1998: Done in `53228623`: dropped the `struct` prefixes and included the protocol headers for the short names.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in `f8a1b66e`: `locate_handle()` now formats/logs once before the singleton branch and keeps the singleton buffer handling below it.

lib/uefi/uefi_platform.cpp
zhangxp1998: Done in `53228623`: `select_device_trees()` now selects the first base device tree and returns `EFI_STATUS_NOT_FOUND` if none is present.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in `f8a1b66e`: moved `fmt_guid()`/`name_buf` after the singleton-handle fast path in `locate_protocol()`.

lib/uefi/boot_service_provider.cpp
zhangxp1998: Done in `f8a1b66e`: removed the explicit `EFI_TEXT_INPUT_PROTOCOL_GUID` check; it now falls through to `EFI_STATUS_NOT_FOUND`.

lib/uefi/uefi_platform.cpp
```diff
@@ -94,7 +97,25 @@ __WEAK EfiStatus select_device_trees(struct GblEfiOsConfigurationProtocol* self,
...
+    for (size_t i = 0; i < num_device_trees; i++) {
+      device_trees[i].selected = false;
+      if (selected_device_tree == num_device_trees &&
+          device_trees[i].metadata.type ==
+              GBL_EFI_DEVICE_TREE_TYPE_DEVICE_TREE) {
+        selected_device_tree = i;
+      }
+    }
+    if (selected_device_tree == num_device_trees) {
+      return EFI_STATUS_NOT_FOUND;
```
silverneko: According to https://android.googlesource.com/platform/bootable/libbootloader/+/refs/heads/gbl-mainline/gbl/docs/gbl_efi_os_configuration_protocol.md#gbl_efi_os_configuration_protocol_selectdevicetrees

select_device_trees never return `EFI_STATUS_NOT_FOUND`.

Either return `UNSUPPORTED` to inform GBL to autoselect, or return `INVALID_PARAMETER` to inform GBL that none are enabled and boot would fail.

zhangxp1998: Done in `532068e3`: fallback now returns `EFI_STATUS_UNSUPPORTED` when no base device tree is found, so GBL can autoselect instead of seeing the unsupported `EFI_STATUS_NOT_FOUND`.

lib/uefi/uefi_platform.cpp
zhangxp1998: Follow-up: after the later return-status comment, the no-base fallback now returns `EFI_STATUS_UNSUPPORTED` in `532068e3`. The selection logic still clears existing selections and selects the first base device tree.

lib/uefi/uefi_platform.cpp
```diff
@@ -94,7 +97,33 @@ __WEAK EfiStatus select_device_trees(struct GblEfiOsConfigurationProtocol* self,
...
+        selected_overlay = i;
+      }
+    }
+    if (selected_device_tree == num_device_trees) {
+      return EFI_STATUS_UNSUPPORTED;
+    }
+    device_trees[selected_device_tree].selected = true;
+    if (selected_overlay != num_device_trees) {
+      device_trees[selected_overlay].selected = true;
+    }
```
silverneko: why do we default to select the first overlay?

I think DT overlays should be selected by some criteria, like platform version / SOC name etc. Since we are not checking those, does it make more sense to select none overlays by default? or select ALL by default?

zhangxp1998: The intent here is that this weak `uefi_platform.cpp` implementation is a QEMU-like reference implementation rather than a platform policy recommendation. QEMU normally does not provide overlay DTBs, so the exact overlay choice does not matter for the QEMU case.

For real OEM platforms, this hook is the place to filter and locate overlay DTBs from the input list using platform-specific criteria such as board/SOC/version. Selecting the first overlay here is just a simple reference behavior that shows where overlay selection happens; platform code should override it when the input can contain multiple meaningful overlays.

zhangxp1998: Done in `2eee933a`: simplified the selection logic with `select_first_dtb(type)`, then use it to select the first base DTB and the first overlay separately after clearing existing selections.

silverneko: ACK

silverneko (approved):

PR #517, opened 2026-07-02
[app][inetsrv] use correct buffer size in tcp_read

TzengKyle:
## Summary

Fix the buffer size argument passed to `tcp_read()` in `app/inetsrv/inetsrv.c`.

The previous code used `sizeof(buf)`, which evaluates to the size of the
pointer rather than the intended buffer capacity. Use `ECHO_BUFSIZE` instead.

## Testing

- `make qemu-virt-arm64-test`


travisg (approved):
good catch!

PR #520, opened 2026-07-12
[arch][arm64] collapse repeated vbar hang slots into a macro

inc1de:
Factor the fifteen identical wfe; b .-4 loops in the trampoline vector table into a single VBAR_HANG macro.


travisg (approved):
looks good to me! thanks!

PR #521, opened 2026-07-27
Fix reading 32-bit start_cluster instead of only reading 16-bit.

tsungyeh:
For FAT32, the First Cluster Number is stored in the Directory Entry's DIR_FstClusHI (0x14) and DIR_FstClusLO (0x1a). However, target_cluster currently reads only the lower 16 bits, which will cause issues on FAT32 systems when files are not located in low clusters.

### DIFF
--- a/lib/fs/fat/dir.cpp
+++ b/lib/fs/fat/dir.cpp
@@ -480,7 +480,11 @@ status_t fat_find_next_entry(fat_fs *fat, file_block_iterator &dbi, uint32_t &of
             LTRACEF("found filename '%s'\n", *out_filename);
 
             // fill out the passed in dir entry and exit
-            uint16_t target_cluster = fat_read16(ent, 0x1a);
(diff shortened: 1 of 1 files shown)


lib/fs/fat/dir.cpp
travisg: Looks like there's an equivalent issue here, we should fix this as well.

travisg (changes requested):
Thanks for the patch, I think there's another place where we can do it to be in sync, otherwise it looks good!

lib/fs/fat/dir.cpp
tsungyeh: Hi,

Thank you for pointing it out.
We are using an older version, so it didn't have this part.
I have amended the patch.
Thank you very much for taking the time to look into this! 
(And thank you for the Little Kernel too 🙂)

Best Regards,
John

travisg: Thanks!

PR #532, opened 2026-08-29
[lib][uefi] correctness, resource lifetime, and PE parsing fixes

zhangxp1998:
Correctness and resource-lifetime fixes in lib/uefi.

Service tables and loader behavior:

* **leave unimplemented service table slots null**: the boot/runtime service tables were pre-filled with an incrementing byte pattern, so calling an unimplemented service jumped to a garbage address. With value-initialized tables the unimplemented slots stay null, which applications can test for, and a stray call faults at address zero instead of a random location. The existing `set_timer` stub is now actually installed in the table (it was never wired up).
* **make synchronous block writes report unsupported**: `write_blocks`/`flush_blocks` returned `EFI_STATUS_SUCCESS` without touching the device, so applications believed their writes had been persisted. They now return `EFI_STATUS_UNSUPPORTED`, matching the async path (`write_blocks_ex`).
* **abort the load when relocation fails**: `load_sections_and_execute()` ignored `relocate_image()`'s result and jumped to the entry point of a partially relocated image.
* **bound PE header parsing to the bytes actually read**: the parser only checked `e_lfanew` against the file header size; a truncated or malformed image could place the optional header or the section table beyond the bytes the reader returned (parsing uninitialized heap) or beyond the 4 KiB buffer entirely. Every region is now checked against the bytes actually read, and a non-EFI-application subsystem is rejected instead of only warned about.

Variable services:

* **follow the spec in get_variable**: report the required size when returning `EFI_STATUS_BUFFER_TOO_SMALL`, return `EFI_STATUS_INVALID_PARAMETER` instead of copying through a null `Data` pointer, honor `*DataSize` for the synthesized SecureBoot/SetupMode answers instead of writing a byte unconditionally, and report attributes for them.
* **tighten set_variable parameter checks and deletion**: reject a null `VendorGuid` (the store would otherwise match a same-named variable under any GUID) and a non-zero `DataSize` with null `Data`; per spec, a zero `DataSize` now deletes the variable instead of leaving an empty one behind. Also fixes the inverted comment/error message (the code accepts only volatile variables).

Resource lifetime:

* **reuse and release boot memory protocol buffers**: `get_boot_buffer` allocated a fresh buffer on every call and nothing ever released them, so an app that returned to the shell (e.g. a fastboot session that never boots) leaked up to 259 MiB of contiguous pages per run. One buffer per type is now cached, repeated requests get the same buffer, and everything is released at run teardown.
* **close block devices when a run ends**: `open_block_device`, `open_async_block_device`, and `open_efi_erase_block_protocol` called `bio_open` with no matching `bio_close`, leaking a device reference per protocol open per run. Opens are now tracked and closed at run teardown, and a missing device fails cleanly with `EFI_STATUS_NOT_FOUND` instead of dereferencing a null bdev.
* **implement calculate_crc32**: wired to lib/cksum's CRC-32 (already used for the EFI system table pointer checksum) instead of returning `EFI_STATUS_UNSUPPORTED`.

Validation (qemu-virt-arm64-test, Clang/LLD, WERROR=1):

* `helloworld_aa64.efi` loads and runs (return code 0), `uefi_set_var`/`uefi_list_var` behave as before, `ut all` passes 36/36, no faults in the log
* Also validated with #530 and #531 applied on top: two consecutive `uefi_load` runs both succeed, PMM `free_count` identical before/after, `ut all` 36/36

### DIFF
--- a/lib/uefi/blockio2_protocols.cpp
+++ b/lib/uefi/blockio2_protocols.cpp
@@ -27,6 +27,7 @@
 #include <uefi/protocols/block_io2_protocol.h>
 #include <uefi/types.h>
 
+#include "blockio_protocols.h"
 #include "events.h"
 #include "io_stack.h"
 #include "memory_protocols.h"
@@ -134,10 +135,21 @@ EfiStatus flush_blocks_ex(EfiBlockIo2Protocol* self, EfiBlockIo2Token* token) {
 }  // namespace
 
 __WEAK EfiStatus open_async_block_device(EfiHandle handle, const void** intf) {
-  auto dev = bio_open(reinterpret_cast<const char*>(handle));
+  bdev_t *dev = nullptr;
+  EfiStatus status =
+      open_tracked_bdev(reinterpret_cast<const char*>(handle), &dev);
+  if (status != EFI_STATUS_SUCCESS) {
+    if (status == EFI_STATUS_NOT_FOUND) {
+      printf("%s: no such block device\n", __FUNCTION__);
+    }
+    return status;
+  }
   printf("%s(%s)\n", __FUNCTION__, dev->name);
   auto interface = reinterpret_cast<EfiBlockIo2Interface*>(
       uefi_malloc(sizeof(EfiBlockIo2Interface)));
+  if (interface == nullptr) {
+    return EFI_STATUS_OUT_OF_RESOURCES;
+  }
   auto protocol = &interface->protocol;
   auto media = &interface->media;
   protocol->media = media;
--- a/lib/uefi/blockio_protocols.cpp
+++ b/lib/uefi/blockio_protocols.cpp
@@ -60,13 +60,13 @@ EfiStatus read_blocks(EfiBlockIoProtocol *self, uint32_t media_id, uint64_t lba,
 
 EfiStatus write_blocks(EfiBlockIoProtocol *self, uint32_t media_id,
                        uint64_t lba, size_t buffer_size, void *buffer) {
-  printf("%s is called\n", __FUNCTION__);
-  return EFI_STATUS_SUCCESS;
+  printf("Writing blocks from UEFI app is currently not supported to protect "
+         "the device.\n");
+  return EFI_STATUS_UNSUPPORTED;
 }
 
 EfiStatus flush_blocks(EfiBlockIoProtocol *self) {
-  printf("%s is called\n", __FUNCTION__);
-  return EFI_STATUS_SUCCESS;
+  return EFI_STATUS_UNSUPPORTED;
 }
 
 EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
--- a/lib/uefi/blockio_protocols.cpp
+++ b/lib/uefi/blockio_protocols.cpp
@@ -18,6 +18,7 @@
 
 #include <kernel/vm.h>
 #include <lib/bio.h>
+#include <malloc.h>
 #include <string.h>
 #include <uefi/protocols/block_io_protocol.h>
 #include <uefi/types.h>
@@ -75,6 +76,48 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
 }
 }  // namespace
 
+namespace {
+// Every bio_open() for the current image run is tracked so its reference can be
+// released at teardown by close_tracked_bdevs(). This is a growable list rather
+// than a fixed array on purpose: close_protocol() does not release individual
+// Block I/O references, so the tracker counts cumulative opens, and a bounded
+// table would reject opens after enough OpenProtocol/CloseProtocol cycles.
+struct TrackedBdev {
+  bdev_t *dev;
+  TrackedBdev *next;
+};
+TrackedBdev *tracked_bdevs = nullptr;
+}  // namespace
+
+EfiStatus open_tracked_bdev(const char *name, bdev_t **out_dev) {
+  *out_dev = nullptr;
+  bdev_t *dev = bio_open(name);
+  if (dev == nullptr) {
+    return EFI_STATUS_NOT_FOUND;
+  }
+  auto *node = reinterpret_cast<TrackedBdev *>(malloc(sizeof(TrackedBdev)));
+  if (node == nullptr) {
+    // Can't remember the reference, so don't hand out an untracked one.
+    printf("%s: out of memory tracking %s\n", __FUNCTION__, name);
+    bio_close(dev);
+    return EFI_STATUS_OUT_OF_RESOURCES;
+  }
+  node->dev = dev;
+  node->next = tracked_bdevs;
+  tracked_bdevs = node;
+  *out_dev = dev;
+  return EFI_STATUS_SUCCESS;
+}
+
+void close_tracked_bdevs() {
+  while (tracked_bdevs != nullptr) {
+    TrackedBdev *node = tracked_bdevs;
+    tracked_bdevs = node->next;
+    bio_close(node->dev);
+    free(node);
+  }
+}
+
 __WEAK EfiStatus open_block_device(EfiHandle handle, const void** intf) {
   printf("%s(%p)\n", __FUNCTION__, handle);
   auto io_stack = get_io_stack();
@@ -87,7 +130,15 @@ __WEAK EfiStatus open_block_device(EfiHandle handle, const void** intf) {
     return EFI_STATUS_OUT_OF_RESOURCES;
   }
   memset(interface, 0, sizeof(EfiBlockIoInterface));
-  auto dev = bio_open(reinterpret_cast<const char *>(handle));
+  bdev_t *dev = nullptr;
+  EfiStatus status =
+      open_tracked_bdev(reinterpret_cast<const char *>(handle), &dev);
+  if (status != EFI_STATUS_SUCCESS) {
+    if (status == EFI_STATUS_NOT_FOUND) {
+      printf("%s: no such block device\n", __FUNCTION__);
+    }
+    return status;
+  }
   interface->dev = dev;
   interface->protocol.revision = EFI_BLOCK_IO_PROTOCOL_REVISION;
   interface->protocol.reset = reset;
--- a/lib/uefi/blockio_protocols.cpp
+++ b/lib/uefi/blockio_protocols.cpp
@@ -109,6 +109,19 @@ EfiStatus open_tracked_bdev(const char *name, bdev_t **out_dev) {
   return EFI_STATUS_SUCCESS;
 }
 
+bool close_tracked_bdev(const char *name) {
+  for (TrackedBdev **pp = &tracked_bdevs; *pp != nullptr; pp = &(*pp)->next) {
+    if (strcmp((*pp)->dev->name, name) == 0) {
+      TrackedBdev *node = *pp;
+      *pp = node->next;
+      bio_close(node->dev);
+      free(node);
+      return true;
+    }
+  }
+  return false;
+}
+
 void close_tracked_bdevs() {
   while (tracked_bdevs != nullptr) {
     TrackedBdev *node = tracked_bdevs;
--- a/lib/uefi/blockio_protocols.h
+++ b/lib/uefi/blockio_protocols.h
@@ -1,10 +1,19 @@
 #ifndef __LIB_UEFI_BLOCKIO_PROTOCOL_H_
 #define __LIB_UEFI_BLOCKIO_PROTOCOL_H_
 
+#include <lib/bio.h>
 #include <uefi/types.h>
 
 EfiStatus open_block_device(EfiHandle handle, const void **intf);
 
 EfiStatus list_block_devices(size_t *num_handles, EfiHandle *buf);
+
+// bio_open wrapper that remembers the device so the reference can be
+// released by close_tracked_bdevs() when the current image run tears down.
+// On success *out_dev holds the opened device. Returns EFI_STATUS_NOT_FOUND
+// when the device does not exist and EFI_STATUS_OUT_OF_RESOURCES when the
+// tracking table is full (so callers can distinguish the two).
+EfiStatus open_tracked_bdev(const char *name, bdev_t **out_dev);
+void close_tracked_bdevs();
 
 #endif
--- a/lib/uefi/blockio_protocols.h
+++ b/lib/uefi/blockio_protocols.h
@@ -8,12 +8,17 @@ EfiStatus open_block_device(EfiHandle handle, const void **intf);
 
 EfiStatus list_block_devices(size_t *num_handles, EfiHandle *buf);
 
-// bio_open wrapper that remembers the device so the reference can be
-// released by close_tracked_bdevs() when the current image run tears down.
+// bio_open wrapper that remembers the device so the reference can be released
+// at teardown by close_tracked_bdevs(), or individually by close_tracked_bdev()
+// when the app calls CloseProtocol.
 // On success *out_dev holds the opened device. Returns EFI_STATUS_NOT_FOUND
-// when the device does not exist and EFI_STATUS_OUT_OF_RESOURCES when the
-// tracking table is full (so callers can distinguish the two).
+// when the device does not exist and EFI_STATUS_OUT_OF_RESOURCES when a tracking
+// node cannot be allocated (out of memory), so callers can distinguish the two.
 EfiStatus open_tracked_bdev(const char *name, bdev_t **out_dev);
+// Release a single tracked reference to the named device, balancing one open
+// (used by CloseProtocol). Returns true if a tracked open was found and closed;
+// an untracked or duplicate close is a harmless no-op.
+bool close_tracked_bdev(const char *name);
 void close_tracked_bdevs();
 
 #endif
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -17,6 +17,7 @@
 #include "boot_service_provider.h"
 
 #include <endian.h>
+#include <limits.h>
 #include <lib/cksum.h>
 #include <lk/compiler.h>
 #include <stdio.h>
@@ -260,10 +261,22 @@ EfiStatus uninstall_multiple_protocol_interfaces(EfiHandle handle, ...) {
   return EFI_STATUS_UNSUPPORTED;
 }
 EfiStatus calculate_crc32(void *data, size_t len, uint32_t *crc_out) {
-  if (data == nullptr || crc_out == nullptr || len == 0) {
+  if (data == nullptr || crc_out == nullptr) {
     return EFI_STATUS_INVALID_PARAMETER;
   }
-  *crc_out = crc32(0, static_cast<const unsigned char *>(data), len);
+  // A non-null zero-length buffer has a well-defined CRC of 0. crc32() takes an
+  // unsigned int length, so feed large buffers in UINT_MAX-sized chunks rather
+  // than silently truncating; the running crc makes the split transparent.
+  unsigned long crc = 0;
+  const auto *buf = static_cast<const unsigned char *>(data);
+  while (len > 0) {
+    const unsigned int chunk =
+        len > UINT_MAX ? UINT_MAX : static_cast<unsigned int>(len);
+    crc = crc32(crc, buf, chunk);
+    buf += chunk;
+    len -= chunk;
+  }
+  *crc_out = static_cast<uint32_t>(crc);
   return EFI_STATUS_SUCCESS;
 }
 
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -507,6 +507,7 @@ void setup_boot_service_table(EfiBootService *service) {
   service->check_event = switch_stack_wrapper<EfiEvent, check_event>();
   service->create_event = create_event;
   service->close_event = close_event;
+  service->set_timer = set_timer;
   service->stall = stall;
   service->raise_tpl = raise_tpl;
   service->restore_tpl = restore_tpl;
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -17,6 +17,7 @@
 #include "boot_service_provider.h"
 
 #include <endian.h>
+#include <lib/cksum.h>
 #include <lk/compiler.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -258,9 +259,12 @@ EfiStatus uninstall_multiple_protocol_interfaces(EfiHandle handle, ...) {
   printf("%s is unsupported\n", __FUNCTION__);
   return EFI_STATUS_UNSUPPORTED;
 }
-EfiStatus calculate_crc32(void *data, size_t len, uint32_t *crc32) {
-  printf("%s is unsupported\n", __FUNCTION__);
-  return EFI_STATUS_UNSUPPORTED;
+EfiStatus calculate_crc32(void *data, size_t len, uint32_t *crc_out) {
+  if (data == nullptr || crc_out == nullptr || len == 0) {
+    return EFI_STATUS_INVALID_PARAMETER;
+  }
+  *crc_out = crc32(0, static_cast<const unsigned char *>(data), len);
+  return EFI_STATUS_SUCCESS;
 }
 
 EfiStatus uninstall_protocol_interface(EfiHandle handle,
--- a/lib/uefi/boot_service_provider.cpp
+++ b/lib/uefi/boot_service_provider.cpp
@@ -418,6 +418,9 @@ EfiStatus close_protocol(EfiHandle handle, const EfiGuid *protocol,
   } else if (guid_eq(protocol, EFI_DEVICE_PATH_PROTOCOL_GUID)) {
     return EFI_STATUS_SUCCESS;
   } else if (guid_eq(protocol, EFI_BLOCK_IO_PROTOCOL_GUID)) {
+    // Balance the bio reference taken by open_block_device(); teardown still
+    // reclaims any opens the app never closed.
+    close_tracked_bdev(reinterpret_cast<const char *>(handle));
     return EFI_STATUS_SUCCESS;
   } else if (guid_eq(protocol, EFI_DT_FIXUP_PROTOCOL_GUID)) {
     return EFI_STATUS_SUCCESS;
--- a/lib/uefi/memory_protocols.cpp
+++ b/lib/uefi/memory_protocols.cpp
@@ -282,45 +282,41 @@ EfiStatus sync_partition_buffer(struct GblEfiBootMemoryProtocol* self,
   return EFI_STATUS_SUCCESS;
 }
 
+struct BootBufferSlot {
+  GblEfiBootBufferType type;
+  size_t size;
+  size_t align_log2;
+  void* addr;
+};
+
+// 2^21 = 2MB alignment on the load/kernel buffers, required by linux kernel
+BootBufferSlot boot_buffer_slots[] = {
+    {GBL_EFI_BOOT_BUFFER_TYPE_GENERAL_LOAD, 128ul * 1024 * 1024, 21, nullptr},
+    {GBL_EFI_BOOT_BUFFER_TYPE_KERNEL, 64ul * 1024 * 1024, 21, nullptr},
+    {GBL_EFI_BOOT_BUFFER_TYPE_RAMDISK, 64ul * 1024 * 1024, PAGE_SIZE_SHIFT,
+     nullptr},
+    {GBL_EFI_BOOT_BUFFER_TYPE_FDT, 2ul * 1024 * 1024, PAGE_SIZE_SHIFT,
+     nullptr},
+    {GBL_EFI_BOOT_BUFFER_TYPE_PVMFW_DATA, 1024ul * 1024, PAGE_SIZE_SHIFT,
+     nullptr},
+};
+
 EfiStatus get_boot_buffer(struct GblEfiBootMemoryProtocol* self,
                           /* in */ GblEfiBootBufferType buf_type,
                           /* out */ size_t* size,
                           /* out */ void** addr) {
-  if (buf_type == GBL_EFI_BOOT_BUFFER_TYPE_GENERAL_LOAD) {
-    *size = 128ul * 1024 * 1024;
-    // 2^21 = 2MB alignment, required by linux kernel
-    *addr = alloc_page(*size, 21);
-    if (*addr == nullptr) {
+  for (auto& slot : boot_buffer_slots) {
+    if (slot.type != buf_type) {
+      continue;
+    }
+    if (slot.addr == nullptr) {
+      slot.addr = alloc_page(slot.size, slot.align_log2);
+    }
+    if (slot.addr == nullptr) {
       return EFI_STATUS_OUT_OF_RESOURCES;
     }
-    return EFI_STATUS_SUCCESS;
-  } else if (buf_type == GBL_EFI_BOOT_BUFFER_TYPE_KERNEL) {
-    *size = 64ul * 1024 * 1024;
-    *addr = alloc_page(*size, 21);
-    if (*addr == nullptr) {
-      return EFI_STATUS_OUT_OF_RESOURCES;
-    }
-    return EFI_STATUS_SUCCESS;
-  } else if (buf_type == GBL_EFI_BOOT_BUFFER_TYPE_RAMDISK) {
-    *size = 64ul * 1024 * 1024;
-    *addr = alloc_page(*size, PAGE_SIZE_SHIFT);
-    if (*addr == nullptr) {
-      return EFI_STATUS_OUT_OF_RESOURCES;
-    }
-    return EFI_STATUS_SUCCESS;
-  } else if (buf_type == GBL_EFI_BOOT_BUFFER_TYPE_FDT) {
-    *size = 2ul * 1024 * 1024;
-    *addr = alloc_page(*size, PAGE_SIZE_SHIFT);
-    if (*addr == nullptr) {
-      return EFI_STATUS_OUT_OF_RESOURCES;
-    }
-    return EFI_STATUS_SUCCESS;
-  } else if (buf_type == GBL_EFI_BOOT_BUFFER_TYPE_PVMFW_DATA) {
-    *size = 1024ul * 1024;
-    *addr = alloc_page(*size, PAGE_SIZE_SHIFT);
-    if (*addr == nullptr) {
-      return EFI_STATUS_OUT_OF_RESOURCES;
-    }
+    *size = slot.size;
+    *addr = slot.addr;
     return EFI_STATUS_SUCCESS;
   }
   printf("get_boot_buffer(%d, %zu) unsupported\n", buf_type, *size);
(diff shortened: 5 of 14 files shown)


lib/uefi/runtime_service_provider.cpp
```diff
@@ -83,7 +91,7 @@ EfiStatus SetVariable(const uint16_t* VariableName, const EfiGuid* VendorGuid,
     return EFI_STATUS_INVALID_PARAMETER;
   }
```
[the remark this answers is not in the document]
zhangxp1998: Both addressed. `set_variable` now rejects a null `VendorGuid` with `EFI_STATUS_INVALID_PARAMETER` before the store can do a name-only match, and rejects `DataSize > 0` with `Data == nullptr` before anything is copied. The checks live in the runtime-service wrapper (the EFI-facing surface), mirroring what `get_variable` already did for its parameters.

lib/uefi/uefi.cpp
```diff
@@ -170,7 +163,10 @@ int load_sections_and_execute(ImageReader *reader,
   }
   printf("Relocating image from 0x%llx to %p\n", optional_header->ImageBase,
          image_base);
-  relocate_image(image_base);
+  if (relocate_image(image_base) != 0) {
+    printf("Failed to relocate image\n");
+    return ERR_BAD_STATE;
```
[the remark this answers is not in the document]
zhangxp1998: Good catch — the guard was toothless for that case. `relocate_image()` now returns -1 on `EFI_IMAGE_REL_BASED_ARM_MOV32A` instead of logging and breaking, so an image carrying an unsupported ARM MOV32 relocation is rejected before it can reach its entry point with the fixup unapplied. Every other unrecognized type already fell through to `default: return -1`, so all unsupported relocations now abort the load consistently and the `relocate_image() != 0` check actually fires. (Implementing the ARM MOVW/MOVT encoding isn't worth it here — the images we load are aarch64/x86 and use DIR64/HIGHLOW.)

lib/uefi/runtime_service_provider.cpp
```diff
@@ -42,27 +42,35 @@ EfiStatus GetVariable(const uint16_t *VariableName, const EfiGuid *VendorGuid,
...
   buffer[i] = 0;
   if (strncmp(buffer, kSecureBoot, sizeof(kSecureBoot)) == 0 || strcmp(buffer, "SetupMode") == 0) {
-    if (DataSize) {
+    if (*DataSize < 1) {
       *DataSize = 1;
+      return EFI_STATUS_BUFFER_TOO_SMALL;
```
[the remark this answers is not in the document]
zhangxp1998: Fixed. The attributes are now published as soon as the variable is recognized, before the buffer-size check, so a size probe that returns `EFI_STATUS_BUFFER_TOO_SMALL` reports the same `Attributes` as a successful read — matching the stored-variable path, which fills them via `efi_get_variable` before the size comparison. SecureBoot/SetupMode callers now observe consistent attributes regardless of buffer size.

lib/uefi/memory_protocols.cpp
```diff
@@ -321,6 +317,15 @@ EfiStatus get_boot_buffer(struct GblEfiBootMemoryProtocol* self,
 
 }  // namespace
 
+void release_boot_buffers() {
+  for (auto& slot : boot_buffer_slots) {
+    if (slot.addr != nullptr) {
+      free_pages(slot.addr, slot.size / PAGE_SIZE);
+      slot.addr = nullptr;
```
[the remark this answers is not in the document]
zhangxp1998: Both addressed. #530 is now merged, so I rebased this branch on top of it: `free_pages()` maps the physical address back through `paddr_to_kvaddr()` before handing it to `pmm_free_kpages()`, so the boot buffers are genuinely returned to the PMM. `release_boot_buffers()` now also checks the result and only clears `slot.addr` on `EFI_STATUS_SUCCESS`, so a failed free can no longer downgrade a tracked buffer into an untracked leak. Verified end-to-end: two back-to-back `uefi_load` runs leave the PMM `free_count` unchanged (129684 → 129684 → 129684).

lib/uefi/blockio_protocols.cpp
```diff
@@ -75,6 +75,37 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
...
+  for (auto &slot : tracked_bdevs) {
+    if (slot == nullptr) {
+      slot = dev;
+      return dev;
+    }
+  }
+  printf("%s: no slot left to track %s, device reference will not be "
+         "released\n",
+         __FUNCTION__, name);
+  return dev;
```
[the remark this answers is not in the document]
zhangxp1998: Fixed by refusing the open instead of leaking. When every tracking slot is in use, `open_tracked_bdev()` now `bio_close()`s the freshly-opened device and returns `nullptr` (the caller surfaces it as a failed open) rather than returning an untracked reference. Repeatedly requesting a block protocol can therefore no longer leak references past teardown — the count is bounded and every tracked open is closed when the run ends. (Kept the fixed-capacity tracker; failing a 17th concurrent open is acceptable here and simpler than a growable store.)

lib/uefi/runtime_service_provider.cpp
```diff
@@ -42,27 +42,38 @@ EfiStatus GetVariable(const uint16_t *VariableName, const EfiGuid *VendorGuid,
   }
   buffer[i] = 0;
   if (strncmp(buffer, kSecureBoot, sizeof(kSecureBoot)) == 0 || strcmp(buffer, "SetupMode") == 0) {
```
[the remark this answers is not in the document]
zhangxp1998: Fixed. The `SecureBoot`/`SetupMode` branch now also requires `VendorGuid == EFI_GLOBAL_VARIABLE_GUID` (a `memcmp` against the constant in `variable_mem.h`) before synthesizing. A query for either name under any other GUID now falls through to the variable store instead of getting the synthesized value and attributes.

lib/uefi/uefi.cpp
```diff
@@ -321,10 +320,27 @@ int load_pe_file(ImageReader *reader) {
            file_header->SizeOfOptionalHeader, sizeof(IMAGE_OPTIONAL_HEADER64));
     return ERR_BAD_STATE;
   }
+  const size_t nt_headers_end = static_cast<size_t>(dos_header->e_lfanew) +
+                                sizeof(IMAGE_FILE_HEADER) +
+                                file_header->SizeOfOptionalHeader;
```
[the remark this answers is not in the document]
zhangxp1998: Handled in `relocate_image()` before it indexes the directory. It now bails out (treating the image as having no relocations) unless the optional header both declares and physically contains that entry:

```
constexpr size_t kRelocDirEnd =
    offsetof(IMAGE_OPTIONAL_HEADER64, DataDirectory) +
    (IMAGE_DIRECTORY_ENTRY_BASERELOC + 1) * sizeof(IMAGE_DATA_DIRECTORY);
if (optional_header->NumberOfRvaAndSizes <= IMAGE_DIRECTORY_ENTRY_BASERELOC ||
    pe_header->FileHeader.SizeOfOptionalHeader < kRelocDirEnd) {
  printf("No base relocation directory present\n");
  return 0;
}
```

So a short optional header (the 112-byte minimum) can no longer make relocation read section-table bytes as `DataDirectory[5]`.

lib/uefi/memory_protocols.cpp
```diff
@@ -329,6 +325,22 @@ EfiStatus get_boot_buffer(struct GblEfiBootMemoryProtocol* self,
...
+  for (auto& slot : boot_buffer_slots) {
+    if (slot.addr == nullptr) {
+      continue;
+    }
+    // Only drop the pointer once the pages are actually returned; otherwise a
+    // failed free would turn a tracked buffer into an untracked leak.
+    if (free_pages(slot.addr, slot.size / PAGE_SIZE) != EFI_STATUS_SUCCESS) {
+      printf("release_boot_buffers: failed to free %p (%zu bytes)\n", slot.addr,
+             slot.size);
+      continue;
```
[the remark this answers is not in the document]
zhangxp1998: Good point — keeping the pointer was the wrong half to fix. Since `release_boot_buffers()` runs immediately before `reset_heap()` (which tears down the boot aspace via `vmm_free_aspace`), any retained `slot.addr` is unmapped on the next run. `release_boot_buffers()` now always clears the slot, so `get_boot_buffer()` re-allocates and re-maps a fresh buffer next time; a `free_pages()` failure is logged (physical pages leak) rather than left reusable. With #530 merged the common path frees cleanly, so this is just the defensive tail. This supersedes the earlier "only clear on success" change, which is what opened the unsafe-reuse window you spotted.

lib/uefi/blockio_protocols.cpp
```diff
@@ -75,6 +75,39 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
...
+      slot = dev;
+      return dev;
+    }
+  }
+  // Every tracking slot is in use. Returning this reference would leak it at
+  // teardown, so close it and fail the open instead.
+  printf("%s: no free tracking slot for %s, refusing to leak the reference\n",
+         __FUNCTION__, name);
+  bio_close(dev);
+  return nullptr;
```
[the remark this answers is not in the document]
zhangxp1998: Done. `open_tracked_bdev()` now returns an `EfiStatus` and yields the device through an out-param: `EFI_STATUS_NOT_FOUND` when `bio_open()` fails, `EFI_STATUS_OUT_OF_RESOURCES` when the tracker is full (device closed), `EFI_STATUS_SUCCESS` otherwise. All three callers (`open_block_device`, `open_async_block_device`, `open_efi_erase_block_protocol`) propagate that status and only print "no such block device" on `NOT_FOUND`, so tracker exhaustion now surfaces as out-of-resources instead of a bogus not-found.

lib/uefi/variable_mem.cpp
```diff
@@ -87,6 +87,16 @@ void efi_set_variable(const char16_t *variable_name,
...
   struct EfiVariable *var = search_existing_variable(variable_name, guid);
 
+  /* setting a variable with zero size deletes it */
+  if (data_len == 0) {
+    if (var) {
+      list_delete(&var->node);
+      free(var->Data);
+      free(var);
+    }
+    return;
```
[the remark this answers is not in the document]
zhangxp1998: Fixed. `efi_set_variable()` now treats `DataSize == 0` as a delete only when `EFI_VARIABLE_APPEND_WRITE` is clear; a zero-length append is a no-op that leaves the existing value intact. Added the missing attribute constants to `variable_mem.h` (`EFI_VARIABLE_APPEND_WRITE = 0x40`, plus the two authenticated-write bits `0x10`/`0x20`).

lib/uefi/relocation.cpp
```diff
@@ -11,6 +11,19 @@ int relocate_image(char *image) {
...
+  // (SizeOfOptionalHeader) that entry. Without this guard a malformed image
+  // could make us read section-table bytes past the optional header and
+  // mis-parse them as a relocation directory.
+  constexpr size_t kRelocDirEnd =
+      offsetof(IMAGE_OPTIONAL_HEADER64, DataDirectory) +
+      (IMAGE_DIRECTORY_ENTRY_BASERELOC + 1) * sizeof(IMAGE_DATA_DIRECTORY);
+  if (optional_header->NumberOfRvaAndSizes <= IMAGE_DIRECTORY_ENTRY_BASERELOC ||
+      pe_header->FileHeader.SizeOfOptionalHeader < kRelocDirEnd) {
+    printf("No base relocation directory present\n");
+    return 0;
```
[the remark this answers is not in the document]
zhangxp1998: Good catch. `relocate_image()` now computes the load adjustment up front and, when there is no usable relocation directory (absent or empty), only returns success if the image can actually run there: it returns an error whenever the load address differs from `ImageBase` **and** the image is marked `IMAGE_FILE_RELOCS_STRIPPED` (it declared it must load at its preferred base).

I deliberately did not error on the strip-clear case: an image with no relocations and RELOCS_STRIPPED clear is position-independent and safe off base. That is exactly what the bundled `helloworld_aa64.efi` is (ImageBase 0x140000000, loaded at 0x53200000, empty .reloc, Characteristics 0x22) — erroring unconditionally would reject it and other valid PIE images. The RELOCS_STRIPPED flag is the spec-defined signal that an image requires its ImageBase, so keying off it catches the unsafe case (stripped + relocated) without breaking PIE loads. Verified: helloworld still loads off base and runs. Added the `IMAGE_FILE_RELOCS_STRIPPED` constant to `pe.h`.

lib/uefi/blockio_protocols.cpp
```diff
@@ -75,6 +75,41 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
...
+  *out_dev = nullptr;
+  bdev_t *dev = bio_open(name);
+  if (dev == nullptr) {
+    return EFI_STATUS_NOT_FOUND;
+  }
+  for (auto &slot : tracked_bdevs) {
+    if (slot == nullptr) {
+      slot = dev;
+      *out_dev = dev;
+      return EFI_STATUS_SUCCESS;
```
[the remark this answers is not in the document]
zhangxp1998: Addressed by removing the fixed cumulative cap (the second option you suggested). The tracker is now a growable singly-linked list instead of a 16-entry array, so the number of opens across an image run is no longer bounded; `close_tracked_bdevs()` walks the list at teardown, `bio_close()`-ing and freeing every node. Repeated OpenProtocol/CloseProtocol cycles can therefore no longer exhaust it and make the 17th open fail. A `malloc` failure still surfaces as `EFI_STATUS_OUT_OF_RESOURCES`, which is now a genuine out-of-memory rather than an artificial cap.

lib/uefi/blockio_protocols.h
```diff
@@ -1,10 +1,19 @@
...
 
 EfiStatus open_block_device(EfiHandle handle, const void **intf);
 
 EfiStatus list_block_devices(size_t *num_handles, EfiHandle *buf);
 
+// bio_open wrapper that remembers the device so the reference can be
+// released by close_tracked_bdevs() when the current image run tears down.
+// On success *out_dev holds the opened device. Returns EFI_STATUS_NOT_FOUND
+// when the device does not exist and EFI_STATUS_OUT_OF_RESOURCES when the
+// tracking table is full (so callers can distinguish the two).
```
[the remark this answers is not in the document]
zhangxp1998: Fixed the stale contract comment: with the growable list there is no fixed table, so `EFI_STATUS_OUT_OF_RESOURCES` now documents a failed tracking-node allocation, and `open_tracked_bdev`/`close_tracked_bdev` are described accordingly.

While here I also closed the underlying gap that made the cumulative-open tracking necessary in the first place. `close_protocol(EFI_BLOCK_IO_PROTOCOL_GUID)` previously returned success without releasing the reference, so a tracked open was only ever closed at teardown. It now balances the reference: `close_tracked_bdev()` unlinks a single tracked node matching the device (block I/O handles are the bdev name) and `bio_close()`s it. An app that pairs OpenProtocol/CloseProtocol no longer accumulates references for the whole run; teardown still sweeps anything left open, and unlinking first means it cannot double-close a device that was already released.

Verified: arm64 Clang/LLD build, two `uefi_load` runs return 0 with stable free_count (129684), `ut all` 38/38.

lib/uefi/boot_service_provider.cpp
```diff
@@ -258,9 +259,12 @@ EfiStatus uninstall_multiple_protocol_interfaces(EfiHandle handle, ...) {
...
   return EFI_STATUS_UNSUPPORTED;
 }
-EfiStatus calculate_crc32(void *data, size_t len, uint32_t *crc32) {
-  printf("%s is unsupported\n", __FUNCTION__);
-  return EFI_STATUS_UNSUPPORTED;
+EfiStatus calculate_crc32(void *data, size_t len, uint32_t *crc_out) {
+  if (data == nullptr || crc_out == nullptr || len == 0) {
+    return EFI_STATUS_INVALID_PARAMETER;
+  }
+  *crc_out = crc32(0, static_cast<const unsigned char *>(data), len);
```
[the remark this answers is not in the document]
zhangxp1998: Fixed. calculate_crc32 now rejects only null pointers, returns 0 for a non-null empty buffer, and feeds larger buffers through crc32() in UINT_MAX-sized chunks using the running crc, so lengths above UINT_MAX are no longer truncated. (fd361ad6)

lib/uefi/relocation.cpp
```diff
@@ -11,18 +11,42 @@ int relocate_image(char *image) {
   const auto dos_header = reinterpret_cast<IMAGE_DOS_HEADER *>(image);
   const auto pe_header = dos_header->GetPEHeader();
   const auto optional_header = &pe_header->OptionalHeader;
+  const auto Adjust =
+      reinterpret_cast<size_t>(image - optional_header->ImageBase);
```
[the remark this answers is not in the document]
zhangxp1998: Fixed in 3865331f. The load adjustment is now computed from integer addresses (reinterpret_cast<size_t>(image) - ImageBase) instead of forming the pointer image - ImageBase, which pointed far outside the allocation whenever ImageBase was nonzero.

lib/uefi/relocation.cpp
```diff
@@ -11,18 +11,42 @@ int relocate_image(char *image) {
...
+  const bool have_reloc_dir =
+      optional_header->NumberOfRvaAndSizes > IMAGE_DIRECTORY_ENTRY_BASERELOC &&
+      pe_header->FileHeader.SizeOfOptionalHeader >= kRelocDirEnd;
   const auto reloc_directory =
-      optional_header->DataDirectory[IMAGE_DIRECTORY_ENTRY_BASERELOC];
-  if (reloc_directory.Size == 0) {
-    printf("Relocation section empty\n");
+      have_reloc_dir
+          ? optional_header->DataDirectory[IMAGE_DIRECTORY_ENTRY_BASERELOC]
+          : IMAGE_DATA_DIRECTORY{};
```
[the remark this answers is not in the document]
zhangxp1998: Fixed in 3865331f. relocate_image() now takes the allocation size and validates every attacker-controlled offset before dereferencing: the relocation directory must lie within the image, each block must contain its header and fit inside the directory (zero, short, and oversized SizeOfBlock are rejected), and the write extent of each fixup, sized per relocation type, must fall inside the image. Out-of-range directories, blocks, and fixups are now rejected instead of being read or written. Verified: arm64 Clang/LLD build, uefi_load returns 0 twice with stable free_count, ut all 38/38.

lib/uefi/variable_mem.cpp
```diff
@@ -87,6 +87,18 @@ void efi_set_variable(const char16_t *variable_name,
                       size_t data_len) {
   struct EfiVariable *var = search_existing_variable(variable_name, guid);
 
+  /* A zero-size write deletes the variable, but only when this is not an
+   * append. With EFI_VARIABLE_APPEND_WRITE set, appending zero bytes is a
+   * no-op that must leave any existing value unchanged. */
+  if (data_len == 0) {
+    if ((attribute & EFI_VARIABLE_APPEND_WRITE) == 0 && var) {
```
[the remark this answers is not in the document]
zhangxp1998: Fixed in a4a8d5e5. A non-empty EFI_VARIABLE_APPEND_WRITE now concatenates onto the existing value instead of falling through to the replace path, and the existing value is left intact on allocation failure. The append bit is stripped from the stored attributes since it is a per-call modifier rather than stored state.

lib/uefi/uefi.cpp
```diff
@@ -138,6 +129,8 @@ int load_sections_and_execute(ImageReader *reader,
   }
   setup_heap();
   DEFER { reset_heap(); };
+  DEFER { release_boot_buffers(); };
+  DEFER { close_tracked_bdevs(); };
```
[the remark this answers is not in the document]
zhangxp1998: You are right that this is a real teardown/async race, and I have deliberately left it out of this PR. close_tracked_bdevs() only balances the bio reference count; it does not wait for asynchronous BlockIo2 work. read_blocks_async either queues a driver callback or spawns a detached async_bio thread, and nothing joins those before the image returns and teardown runs bio_close.

Doing this correctly means tracking in-flight asynchronous operations before releasing the reference -- an outstanding-op count on the tracked device that teardown drains (or an explicit cancel path) -- which is a larger change to the async path than the reference-lifetime fixes this PR is about. I would rather land it as a focused follow-up than attach a partial guard here. The synchronous BlockIo path completes its I/O before returning and is unaffected; the pending-async-at-exit window is a known limitation I will address separately.

lib/uefi/blockio_protocols.cpp
```diff
@@ -75,6 +76,83 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
...
 }  // namespace
 
+namespace {
+// Each distinct bdev opened during the current image run is tracked once, with
+// a count of how many UEFI opens are still outstanding. Reusing an already-open
+// device keeps the list bounded by the number of bdevs rather than the
+// cumulative number of OpenProtocol calls. The underlying bio reference is
+// taken on the first open and released by close_tracked_bdev() when the last
+// open closes, or by close_tracked_bdevs() at teardown.
+struct TrackedBdev {
```
silverneko, line 87: have you considered using https://github.com/littlekernel/lk/blob/master/top/include/lk/list.h
instead of implementing your own linked list?

lib/uefi/blockio_protocols.cpp
```diff
@@ -75,6 +76,83 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
...
+  for (TrackedBdev *node = tracked_bdevs; node != nullptr; node = node->next) {
+    if (strcmp(node->dev->name, name) == 0) {
+      return node;
+    }
+  }
+  return nullptr;
+}
+}  // namespace
+
+EfiStatus open_tracked_bdev(const char *name, bdev_t **out_dev) {
```
silverneko, line 105: why do we want to reference track the opened devices ourselves? why not just open a new bdev for each open_protocol(), or reference count the blockio_protocol struct instead?

Also, bio_open / bio_close already does its own reference counting & bookkeeping https://github.com/zhangxp1998/lk/blob/5203dcb1b2468ce60b9dc031d49c305fe3c39133/lib/bio/bio.c#L357

so I think we shouldn't ref-count it again here.

lib/uefi/blockio_protocols.cpp
zhangxp1998: Yes, agreed. I switched the tracker to LK's list implementation in 089be41f: `TrackedBdev` now embeds a `list_node`, and lookup, insertion, removal, and teardown use `list_for_every_entry`, `list_add_tail`, `list_delete`, and `list_remove_head_type`. The one-entry-per-distinct-bdev/open-count behavior is unchanged.

lib/uefi/blockio_protocols.cpp
zhangxp1998: `OpenProtocol()` can be called multiple times for the same handle/protocol, so there can be several outstanding UEFI opens referring to the same bdev. The extra `open_count` is meant to track those logical UEFI opens, not to duplicate BIO's device-lifetime refcount.

The tracker deliberately takes one `bio_open()` reference per distinct bdev and multiplexes repeated UEFI opens onto it. This avoids adding another cleanup-list node and another BIO reference for every repeated `OpenProtocol()` call. `CloseProtocol()` drops one logical open; the single BIO reference is released when the last logical open closes. If the application returns without balancing all of its opens, the per-run teardown still has one entry for the bdev and can release that reference, so an uncooperative or buggy UEFI application cannot leak device references across runs.

BIO's own reference count keeps a bdev alive, but BIO does not know which references belong to the current UEFI image run or when that run ends, so it cannot provide that cleanup by itself. Taking a fresh BIO reference for every `OpenProtocol()` would still require an unbounded per-open cleanup list (or an equivalent count) to recover missing `CloseProtocol()` calls at teardown.

Reference-counting a shared Block I/O protocol object could also model this, but the current UEFI shim allocates transient protocol interfaces from the per-run heap and `CloseProtocol()` receives the handle/GUID rather than the returned interface pointer. Doing that cleanly would require a broader handle/protocol registry. I would prefer to keep this small per-bdev ownership tracker here.

lib/uefi/blockio2_protocols.cpp
```diff
@@ -134,10 +135,21 @@ EfiStatus flush_blocks_ex(EfiBlockIo2Protocol* self, EfiBlockIo2Token* token) {
...
+    if (status == EFI_STATUS_NOT_FOUND) {
+      printf("%s: no such block device\n", __FUNCTION__);
+    }
+    return status;
+  }
   printf("%s(%s)\n", __FUNCTION__, dev->name);
   auto interface = reinterpret_cast<EfiBlockIo2Interface*>(
       uefi_malloc(sizeof(EfiBlockIo2Interface)));
+  if (interface == nullptr) {
+    return EFI_STATUS_OUT_OF_RESOURCES;
```
silverneko, line 158: should we call `close_tracked_bdev` before returning to decrement the counter?

lib/uefi/blockio_protocols.cpp
```diff
@@ -87,7 +166,15 @@ __WEAK EfiStatus open_block_device(EfiHandle handle, const void** intf) {
...
   memset(interface, 0, sizeof(EfiBlockIoInterface));
-  auto dev = bio_open(reinterpret_cast<const char *>(handle));
+  bdev_t *dev = nullptr;
+  EfiStatus status =
+      open_tracked_bdev(reinterpret_cast<const char *>(handle), &dev);
+  if (status != EFI_STATUS_SUCCESS) {
+    if (status == EFI_STATUS_NOT_FOUND) {
+      printf("%s: no such block device\n", __FUNCTION__);
+    }
+    return status;
```
silverneko, line 182: Either free `interface` before returning, or call `uefi_malloc` only if  `open_tracked_bdev` succeeded

lib/uefi/blockio_protocols.cpp
```diff
@@ -75,6 +77,83 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
...
+  list_for_every_entry(&tracked_bdevs, entry, TrackedBdev, node) {
+    if (strcmp(entry->dev->name, name) == 0) {
+      return entry;
+    }
+  }
+  return nullptr;
+}
+}  // namespace
+
+EfiStatus open_tracked_bdev(const char *name, bdev_t **out_dev) {
```
silverneko, line 108: since this can be called from an asynchronous context, I think the entire tracked_bdev list, and critical sections accessing this list, should be protected with mutex to ensure exclusive access from a single execution thread at a time.

lib/uefi/blockio_protocols.cpp
silverneko: ACK

lib/uefi/uefi_platform.cpp
```diff
@@ -377,7 +378,10 @@ __WEAK EfiStatus open_efi_erase_block_protocol(EfiHandle handle, const void** in
     return EFI_STATUS_OUT_OF_RESOURCES;
   }
   memset(p, 0, sizeof(*p));
-  p->dev = bio_open(device_name);
+  EfiStatus status = open_tracked_bdev(device_name, &p->dev);
+  if (status != EFI_STATUS_SUCCESS) {
+    return status;
```
silverneko: Either free interface before returning, or call uefi_malloc only if open_tracked_bdev succeeded

lib/uefi/relocation.cpp
```diff
@@ -7,81 +7,143 @@
 
 #include "pe.h"
 
-int relocate_image(char *image) {
+int relocate_image(char *image, size_t image_size) {
   const auto dos_header = reinterpret_cast<IMAGE_DOS_HEADER *>(image);
   const auto pe_header = dos_header->GetPEHeader();
   const auto optional_header = &pe_header->OptionalHeader;
+  // Compute the load adjustment from integer addresses. Subtracting ImageBase
```
silverneko: IMO this comment is really confusing.

It's not that "forming a pointer far outside the allocation" is UB (whatever "far outside" means???)

It's that pointer arithmetic with pointers from different allocation are UB.

The comment should just say "convert to interger type as pointer arithmetic is UB", or don't write this comment at all.

lib/uefi/relocation.cpp
```diff
@@ -7,81 +7,143 @@
...
+    // its preferred base or is position independent. An image whose relocations
+    // were stripped (IMAGE_FILE_RELOCS_STRIPPED: it must load at ImageBase) that
+    // landed elsewhere would run with unadjusted absolute addresses, so reject
+    // it instead of reporting success.
+    if (Adjust != 0 &&
+        (pe_header->FileHeader.Characteristics & IMAGE_FILE_RELOCS_STRIPPED)) {
+      printf("Image relocations stripped but not loaded at its ImageBase\n");
+      return -1;
+    }
+    printf("%s\n", have_reloc_dir ? "Relocation section empty"
```
silverneko: nitpick: stick with either "relocation section" or "relocation directory" to be consistent

lib/uefi/relocation.cpp
```diff
@@ -7,81 +7,143 @@
...
+                                  : "No base relocation directory present");
     return 0;
   }
+  // The relocation directory is attacker-controlled. Keep the whole region
+  // within the allocated image so a malformed PE cannot walk out of bounds.
+  if (reloc_directory.VirtualAddress > image_size ||
+      reloc_directory.Size > image_size - reloc_directory.VirtualAddress) {
```
silverneko: this may underflow? it's safer to cast `reloc_directory.Size/VirtualAddress` to u64 and check `reloc_directory.Size + reloc_directory.VirtualAddress > image_size` instead so no risk of overflow

lib/uefi/relocation.cpp
```diff
@@ -7,81 +7,143 @@
 
 #include "pe.h"
 
-int relocate_image(char *image) {
+int relocate_image(char *image, size_t image_size) {
```
silverneko: why do we need to pass in an Image_size argument? Shouldn't this be obtainable from the COFF header?

lib/uefi/blockio2_protocols.cpp
zhangxp1998: Good catch. Fixed in ab1b0129. The tracked open is now guarded by `DEFER`, so any failure after `open_tracked_bdev()` automatically calls `close_tracked_bdev()`. The guard is disabled only after the completed interface is returned to the caller.

lib/uefi/blockio_protocols.cpp
zhangxp1998: Fixed in ab1b0129. `open_tracked_bdev()` now runs before `uefi_malloc()`, and a `DEFER` guard rolls the tracked open back on every subsequent failure path. This also keeps future early returns from reintroducing the leak.

lib/uefi/blockio_protocols.cpp
zhangxp1998: Agreed. Fixed in ab1b0129 by adding a `Mutex` for the tracker. Lookup, open-count updates, list insertion/removal, and teardown are now all serialized with `AutoLock`; the internal lookup helper was renamed to `find_tracked_bdev_locked()` to make the locking requirement explicit.

lib/uefi/uefi_platform.cpp
zhangxp1998: Fixed in ab1b0129. The erase-block path now opens the tracked bdev before allocating the interface and installs a `DEFER` rollback immediately after the successful open. The guard is disabled only once the completed interface is handed to the caller.

lib/uefi/relocation.cpp
zhangxp1998: Agreed. Fixed in ab1b0129. The comment now simply says that the adjustment is computed using integers because pointer arithmetic on pointers from different allocations is undefined behavior in C++.

lib/uefi/relocation.cpp
zhangxp1998: Fixed in ab1b0129. The empty case now says "Relocation directory empty", consistently with the PE data-directory terminology used by the surrounding code.

lib/uefi/relocation.cpp
zhangxp1998: Fixed in ab1b0129. Both fields are promoted to `uint64_t` before addition, and the resulting end offset is compared with `image_size`, avoiding subtraction underflow as suggested.

lib/uefi/relocation.cpp
zhangxp1998: Agreed. Fixed in ab1b0129. The comment now simply says that the adjustment is computed using integers because pointer arithmetic on pointers from different allocations is undefined behavior in C++.

lib/uefi/relocation.cpp
zhangxp1998: Fixed in ab1b0129. The empty case now says "Relocation directory empty", consistently with the PE data-directory terminology used by the surrounding code.

lib/uefi/relocation.cpp
zhangxp1998: Fixed in ab1b0129. Both fields are promoted to `uint64_t` before addition, and the resulting end offset is compared with `image_size`, avoiding subtraction underflow as suggested.

silverneko:
BTW I don't have permission to mark comment threads as "resolved", so you'll have to do that once you resolved some comment.

lib/uefi/blockio_protocols.cpp
```diff
@@ -75,19 +79,120 @@ EfiStatus reset(EfiBlockIoProtocol *self, bool extended_verification) {
...
+  AutoLock lock{&tracked_bdevs_mutex};
+  TrackedBdev *entry;
+  while ((entry = list_remove_head_type(&tracked_bdevs, TrackedBdev, node)) !=
+         nullptr) {
+    bio_close(entry->dev);
+    free(entry);
+  }
+}
+
 __WEAK EfiStatus open_block_device(EfiHandle handle, const void** intf) {
```
silverneko, line 169: I found both star to the right/left styles exist in the code base `void* p` vs. `void *p`. We should stick to a single consistent style. If there is no project wide consistent style in the `lk` repo, at least have consistent style in `lib/uefi` subdir.

This doesn't need to be done in this change though. Can be done in a follow up change that purely formats the code.

lib/uefi/boot_service_provider.cpp
```diff
@@ -258,9 +260,24 @@ EfiStatus uninstall_multiple_protocol_interfaces(EfiHandle handle, ...) {
...
+    return EFI_STATUS_INVALID_PARAMETER;
+  }
+  // A non-null zero-length buffer has a well-defined CRC of 0. crc32() takes an
+  // unsigned int length, so feed large buffers in UINT_MAX-sized chunks rather
+  // than silently truncating; the running crc makes the split transparent.
+  unsigned long crc = 0;
+  const auto *buf = static_cast<const unsigned char *>(data);
+  while (len > 0) {
+    const unsigned int chunk =
+        len > UINT_MAX ? UINT_MAX : static_cast<unsigned int>(len);
```
silverneko: use std::min?

```cpp
const auto chunk = std::min<unsigned>(len, UINT_MAX);
```

silverneko (approved):

lib/uefi/boot_service_provider.cpp
zhangxp1998: Switched to `std::min` in a78f6437. I kept the comparison in `size_t` and converted the clamped result to `unsigned int`:

```cpp
const auto chunk = static_cast<unsigned int>(std::min<size_t>(len, UINT_MAX));
```

Using `std::min<unsigned>` would narrow `len` before the comparison, losing the large-buffer handling. This version preserves it while using the standard helper.

lib/uefi/blockio_protocols.cpp
zhangxp1998: Agreed. The root `.clang-format` already specifies `PointerAlignment: Right` (`void *p`), so that is the convention to use consistently in `lib/uefi` as well. I'll keep the directory-wide cleanup in a separate, formatting-only follow-up as suggested, without mixing it into these correctness changes. Leaving this thread open to track that follow-up.

lib/uefi/relocation.cpp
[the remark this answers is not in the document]
zhangxp1998: While rechecking this thread before resolving it, I also tightened the block-size validation in 17c06864: the payload must contain a whole number of 16-bit relocation entries before the block is walked. This complements the existing directory, block-extent, and fixup-extent checks.

The final tree passes the arm64 Clang/LLD build with warnings as errors, two consecutive `uefi_load` runs (both return 0, PMM free count stays 129683), and all 38 unit tests. These are the normal-image/runtime regressions, not a malformed-image test suite.

zhangxp1998:
Thanks for the review and the note about resolving threads. I've rechecked the older discussions against the current tree and marked 27 threads resolved.

The `std::min` change is in a78f6437 (comparison stays in `size_t`), and 17c06864 finishes the relocation block-size check for complete entries.

Two follow-ups are intentionally still open:
- [Pointer formatting consistency](https://github.com/littlekernel/lk/pull/532#discussion_r4013620460): a separate formatting-only cleanup using the existing `PointerAlignment: Right` setting.
- [Draining outstanding asynchronous I/O before teardown](https://github.com/littlekernel/lk/pull/532#discussion_r3902240989): the previously deferred async-lifetime work, not claimed fixed by the tracker mutex.

Current head: 17c06864. Local validation passed: arm64 Clang/LLD with `WERROR=1`, two successful UEFI loads with PMM free count unchanged at 129683, and 38/38 unit tests. CI has been triggered for the new head.

